PR #8, opened 2022-08-19
add 41-48 and condition enum

jonnycodesalot:
Managed to successfully perform db:migrate, so hopefully this is good now.

### DIFF
--- a/backend/db/migrate/20220817011608_create_property_assessments.rb
+++ b/backend/db/migrate/20220817011608_create_property_assessments.rb
@@ -1,8 +1,7 @@
 class CreatePropertyAssessments < ActiveRecord::Migration[7.0]
   def change
     create_table :property_assessments do |t|
-
-      t.timestamps
+      t.timestamps 
       t.string :heat_type
       t.string :heat_fuel
       t.string :ac_type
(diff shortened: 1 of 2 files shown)


backend/db/migrate/20220817011608_create_property_assessments.rb
```diff
@@ -1,8 +1,22 @@
 class CreatePropertyAssessments < ActiveRecord::Migration[7.0]
+  def up
+    execute <<-SQL
+      CREATE TYPE condition AS ENUM ('not_set', 'poor', 'fair', 'average', 'good', 'excellent');
+    SQL
+  end
+  # Type for condition values for parcels, rooms, etc.
+  enum condition:  { not_set: 'not_set', poor: 'poor', fair: 'fair', average:'average',  good: 'good', excellent: 'excellent' }
```
mzagaja, line 8: Let's use the Rails enum as shown in http://api.rubyonrails.org/classes/ActiveRecord/Enum.html for now, so we don't have to worry about pulling through the native Postgres type.

mzagaja (changes requested):
* Update to use Rails enum
* Fix quote/colon syntax on types

mzagaja (approved):
Looking a lot better now! Congrats on your first Code for Boston pull request.

Issue #20, opened 2022-08-25
Setup Continuous Integration with GitHub Actions


mzagaja:
### Overview
In order to facilitate consistency in our code base we want to setup Continuous Integration with GitHub Actions. 

### Action Items
Setup GitHub Actions Continuous Integration in the repo so it will run the backend test suite using RSpec, lint the code using Rubocop, and check for security issues using `bundle audit`. 

### Resources/Instructions
- [GitHub Actions Template for Ruby on Rails Projects](https://gist.github.com/mzagaja/e4ce82c421965e14c29a00d07bbd8cad)

PR #27, opened 2022-08-25
#20 Create GitHub Actions automation

zganger:
This PR adds GitHub actions automation to address issue #20. The stages all pass excluding rubocop, but the complaints it lists are legitimate. I did not address them as I am not a ruby expert, but I would accept guidance and add a revision, or we can consider this working and publish a follow-up PR to address the rubocop concerns.


mzagaja:
I'll try taking a look at this tomorrow. For the balance of complaints rubocop has do you want to run:

```
rubocop --auto-gen-config
```
and commit the result?

This should [create a todo config file](https://docs.rubocop.org/rubocop/configuration.html#automatically-generated-configuration) and allow CI to pass.

zganger:
Thanks, I added the todo file in the latest commit.

.github/workflows/test.yml
```diff
@@ -0,0 +1,59 @@
+name: Test
+
+on:
+  push:
```
jonnycodesalot, line 4: I wonder if we would really want this, from the perspective of spending credits?  Might be adequate to just run it for PRs to main.  But having it on individual pushes on other branches is certainly handy if credits are effectively free.

jonnycodesalot: @mzagaja

jonnycodesalot (approved):
Looks good to me, look forward to leveraging it!

PR #29, opened 2022-08-31
Missing table fields

AlisonPAnderson:
I noticed several table fields were missing (including year built) - so I added them


backend/db/migrate/20220817011608_create_property_assessments.rb
```diff
@@ -51,7 +51,7 @@ def change
       t.string :bthrm_style3
       t.string :kitchen_type
       t.string :kitchen_style1
-      t.string :kitchen_style
+      t.string :kitchen_style  # is this kitchen_style2?
```
cheradenine, line 54: yes, should be kitchen_style2

backend/db/migrate/20220817011608_create_property_assessments.rb
```diff
@@ -63,6 +63,17 @@ def change
       t.integer :kitchens_count
       t.integer :rooms_total_count
 
+    
+      # I noticed these fields were missing from the table.
```
cheradenine, line 67: Nice! You can remove the comment from the final code though.

cheradenine (commented):
This PR has some other unrelated commits in it. Maybe you need to rebase or otherwise base this on main?
If it is ok I'm going to go into this same migration file and re-oder all the fields so that they are the same as the original dataset. And I can take care of adding these missing ones - or I can leave just those out if you want to do the honors?

AlisonPAnderson:
closing per @cheradenine  - his request covers this issue!

PR #30, opened 2022-09-01
Reoders fields to match data set. Normalizes all names to match columns.

cheradenine:
This PR does the following:
* Reorders all the fields to be identical to what is in the source dataset.
* Changes all the names to be consistent with the abbreviations
* * We had some that were modified to be more readable but for data debugging it will be easier if they are the same
* Adds comment to the data key document
* Changes some of the string fields to limit number of chars (for single letter codes)
* Convert values that are in the dataset as 'Y' or 'N' to boolean - will need to handle this conversion on the data import

### DIFF
--- a/backend/db/migrate/20220817011608_create_property_assessments.rb
+++ b/backend/db/migrate/20220817011608_create_property_assessments.rb
@@ -19,7 +19,7 @@ class CreatePropertyAssessments < ActiveRecord::Migration[7.0]
       t.integer :num_bldgs
       t.integer :luc
       t.string  :lu, limit: 2
-      t.string  :luc_desc
+      t.string  :lu_desc
       t.string  :bldg_type
       t.boolean :own_occ # Boolean represented as Y or N
       t.string  :owner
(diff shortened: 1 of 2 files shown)


backend/db/migrate/20220817011608_create_property_assessments.rb
```diff
@@ -1,69 +1,77 @@
...
-
-      t.string :owner
-      t.string :mail_addressee
-      t.string :mail_address
-      t.string :mail_city
-      t.string :mail_state, limit: 2
-      t.string :mail_zipcode # these can be 5 or zip+4 format
-      t.float :res_floor
+      t.string  :lu, limit: 2
+      t.string  :luc_desc
```
jonnycodesalot, line 22:
```suggestion
      t.string  :lu_desc
```

Typo I think - it's lu_desc in the spreadsheet

jonnycodesalot (approved):
Thanks for doing this!  Definitely much easier to digest with them all matching.  I don't like that there are a few fields where the meaning isn't obvious to me from the name, but I don't feel strongly about which is the right way to go there. This definitely is a big improvement regardless :-)

AlisonPAnderson (approved):
Thank you John!  This does help a lot!

jonnycodesalot:
I actually just found I was unable to perform db:migrate with this branch, where I was previously.  Not sure why yet, but i rescinded the approvals :-/

jonnycodesalot (approved):

PR #32, opened 2022-09-07
Seed import

jonnycodesalot:
UPDATE:
- Added config lines to specify using the full csv for production, but the test one for development/test
- Added a rake task to download the full csv from the web (seed:download)
- Added a rake task to load the 1000-record test seed file (seed:test)
- Added a rake task to load the full csv, assuming it's been downloaded (seed:full)

The last two will be redundant vs db:seed, depending on your configuration.  Note that db:reset appears to run the seed task, so not so sure how useful those rake tasks are anymore.

Not sure about all the naming, and I had to trick Rubocop, probably because I don't actually know how to import code across files properly.  Welcome corrections.

===============================================================

Use smarter csv to import the database.  I was able to run this successfully on the full csv (178K records). I did change several fields, so if we merge as-is y'all will need to rollback and re-migrate.

I've also added a sample CSV with the first 1000 records, which people can use to play with data locally.

HELP NEEDED: Not sure how to reference the csv file in the code.  Ideally I'd have some sort of configuration set variable, because it would be nice if the push test would run a seed import on the small 1000 record file, to validate that continues to work.

Some notes:
- Some weird records were encountered that broke some of the limit rules that were previously there (like Canada being listed as a state)
- There's a lot more refinement that could be done to fields that ought to be represented as boolean, like CORNER_UNIT.
- Had to go to bigint to support all values for the ID fields.

Some of the fields are very inconsistently used, and could use more refinement.  The thinking @cheradenine and I had was that it would probably be better to see what fields that data folks actually want to make use of, then refine them with migrations as necessary.  For now we just wanted to get all the data in.


cheradenine (approved):
Looks great.
I think with tests it is possible to have a test database and create records in it, which could give us a way to validate that the 1000 record sample can be imported. But then the test running in github actions would need a database so not sure how to do that.

jonnycodesalot:
> Looks great. I think with tests it is possible to have a test database and create records in it, which could give us a way to validate that the 1000 record sample can be imported. But then the test running in github actions would need a database so not sure how to do that.

Actually it's already spinning up a database - if you click the "all checks have passed" and look at the tests that run, it looks like the setup that Zack put together is spinning up a db and running db:migrate.  So i feel like I should be able to jam it in there, if I just figure out how the filename should be referenced.  Won't be able to play with it again before Tuesday, but maybe I can mess with it before the meeting if anybody can post suggestions.

Also if you think you know how to do it, feel free to just merge an additional change into this branch.  No hard feelings. :)

mzagaja:
For the test database the best practice is to setup a Factory for the model using [FactoryBot](https://github.com/thoughtbot/factory_bot). It just needs to follow the general shape of the sample data without using it directly (tests should exercise the format of the data but not be dependent upon it *most of the time*).  Sometimes [Faker](https://github.com/faker-ruby/faker) can be useful for creating more realistic/varying fake data.

mzagaja (changes requested):
This is mostly good but I'm noticing ZIP codes are importing with the leading 0s removed. `mail_zipcode` comes in as a `String` type but misses the 0:

```
mail_zipcode: "2128",
```

and `zip_code` comes in as an integer, should probably be changed to a `String`:

```
zipcode: 2128,
```

jonnycodesalot:
> This is mostly good but I'm noticing ZIP codes are importing with the leading 0s removed. `mail_zipcode` comes in as a `String` type but misses the 0:
> 
> ```
> mail_zipcode: "2128",
> ```
> 
> and `zip_code` comes in as an integer, should probably be changed to a `String`:
> 
> ```
> zipcode: 2128,
> ```

I fixed the datatype, but John and I had suggested that we handle mutating individual fields with migrations, and i would expect the mail_zipcode field to be part of that.  This particular PR is kind of dragging on, so I'd suggest we merge it so we can collaborate on the individual field update a bit.

PR #44, opened 2022-09-13
Initial devise install and code gem

cheradenine:
Installs the devise gem for handling user accounts, login, sessions etc.
This is pretty much the installation and initialization that sets up a bunch of defaults.


mzagaja (approved):
Should be good to merge in1

PR #52, opened 2022-09-17
Create READMEs

runkelcorey:
Drafted README for project root and created blank subdirectories, each with their own README, for data-analysis, marketing, and partnerships. Pretty much guessed at the tech used, so please add commits as necessary.


mzagaja (approved):
This looks great to me!

PR #85, opened 2022-10-14
Issue 73

jackberrypassionfruit:
Pushed footer to bottom using flex rules (flexGrow: 1 on body components and minHeight: 100vh on container) on Material Ui Boxes


Zevez7:
you need to push this to dat_frontend branch instead of main, please try again.

Issue #78, opened 2022-10-07
Create an account page for the surveyor view


Zevez7:
### Overview
On the surveyor view create a account page that has a mobile and tablet view.
The account page should have these features

### Action Items
1. First Name
2. Last Name
3. Email 
4. Address
5. Phone number

Button: Edit
Button: Request New Assignment

Add a link to his page on the surveyor view nav folder with the left slide menu.

Resources: 
Use the dashboard page as a design guide on how to create the box/grid that will allow mobile and tablet view.
Contact Dat on the slack for more help.

PR #90, opened 2022-10-18
Account Edit page

aTegart:
Related to #78 

### Features:
- Adds `accountSlice` for managing account details
- Account details that are edited in the `EditAccount` form will be shown on the Account page (kept in local page state)
- Editing form has basic validation, requiring a value for all fields, email format validation, and enforcing phone numbers as being 8-11 characters - we should decide which of this validation makes sense

### Future TODO:
- Update route registration for `/account/edit` page to be more idiomatic
- Connect to API
- Use `setLogOut` action to clear local state when user logs out?
- Decide which fields should be mandatory and consider marking them with asterisks
- Update the look-and-feel to be less similar to the login page
- Update any fields that get changed when the surveyor user model is solidified (e.g. Address will likely become multiple parts)
- Auto-format phone numbers to divide the number visually into a `(XXX) XXX-XXXX` format

### Screenshots:
Desktop:
image

iPhone:
image

iPad:
image

### DIFF
--- a/frontend/front/src/features/account/accountSlice.js
+++ b/frontend/front/src/features/account/accountSlice.js
@@ -1,0 +1,34 @@
+import { createSlice } from "@reduxjs/toolkit";
+
+const initialState = {
+  firstName: "John",
+  lastName: "Smith",
+  email: "jsmith@example.com",
+  address: "12345 John Smith Way",
+  phoneNumber: "1234567890",
+};
+
+const accountSlice = createSlice({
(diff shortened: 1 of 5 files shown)


Zevez7:
Approved and merged. I would like you to use sx instead of style for css. Please go back and change all style to sx. SX provide some powerful features that is very useful from material ui. You can add this change to another pull request whenever you want. Low priority. Thanks Dat

PR #102, opened 2022-11-08
Run tests on main

aTegart:
Checking if tests are broken on main


aTegart:
Fixed tests in #44 , no longer needed

PR #113, opened 2022-11-29
Resolve #107 Surveyors endpoint

### LINKED ISSUE #107: Update User Endpoint
Need following attributes for a User:

- id
- firstName
- lastName
- email
- phone
- streetAddress
- Street number and street name together
- geoCode
- city
- zipCode
- state
- role
- admin, surveyor
- status
- active, disabled

Update Controller to comport to the filtering they need.

image

aTegart:
Resolves #107 

Results from `bin/rails generate scaffold Surveyor user:references firstname:string lastname:string email:string phone:string street_address:string geocode:string city:string zipcode:string state:string role:string status:string`

Named `Surveyor` to avoid collision with devise `Users`.

Maintains foreign key reference to devise `user`.

Supports filtering by :city, :zipcode, :state, :role, :status


aTegart:
Failing due to warning:
```Confidence: Medium
Category: Mass Assignment
Check: PermitAttributes
Message: Potentially dangerous key allowed for mass assignment
Code: params.require(:surveyor).permit(:user_id, :firstname, :lastname, :email, :phone, :street_address, :geocode, :city, :zipcode, :state, :role, :status)
File: app/controllers/surveyors_controller.rb
Line: 68
```

Should some of those params (e.g. role) be assignable only indirectly?

mzagaja:
It's not entirely clear to me which key it is refering to here, but my best guess would be `:role`. Understandbly from a security standpoint, you don't want to be able to update/assign roles from a public API endpoint. However I think separating this is a bit beyond the scope of the goal of this PR, so I'm ok with [ignoring that for now](https://brakemanscanner.org/docs/ignoring_false_positives/).

backend/spec/factories/surveyors.rb
```diff
@@ -0,0 +1,16 @@
...
+    lastname { "MyString" }
+    email { "MyString" }
+    phone { "MyString" }
+    street_address { "MyString" }
+    geocode { "MyString" }
+    city { "MyString" }
+    zipcode { "MyString" }
+    state { "MyString" }
+    role { "MyString" }
+    status { "MyString" }
```
mzagaja, line 14: @aTegart Usually before committing this I'll update the sample values here from `MyString` to something that looks like the data we'll be using. You can look at mine for an example.

mzagaja (changes requested):
This looks reasonable to me. The only other major change I would make here is since we won't be using them, you can safely remove all the files ending in `.html.erb` and then the associated tests with 

```sh
rm spec/views/surveyors/*.html.erb_spec.rb
rm app/views/surveyors/*.html.erb
```

mzagaja (approved):
🚀

PR #116, opened 2022-12-07
update docker image to use alpine linux for size & speed improvements

zganger:
Utilize alpine linux to reduce size and improve performance of backend application.

- image size reduced from 420mb to 150mb.
- reduced vulnerabilities in image to 0 from 
   ```
     1 critical
     19 high
     93 medium
     22 low
     186 informational
     92 undefined
   ```
- ARM support
- Ruby 3.1.3
- Server runs on AWS ECS


backend/Dockerfile
```diff
@@ -1,15 +1,14 @@
...
-RUN apt-get update -y && apt-get upgrade -y && apt-get install -y postgresql-client
-# for alpine; when dependency resolution is fixed.
-# FROM ruby:3.1.2-alpine
-# RUN apk update && apk upgrade && apk add postgresql-client
+FROM ruby:3.1.3-alpine
+RUN apk update && apk upgrade && apk add build-base libpq-dev postgresql-client gcompat tzdata
 WORKDIR /usr/src/app
 COPY Gemfile Gemfile.lock ./
+# add x86_64 for support on ARM platform - nokogiri expects x86 for distribution
+RUN bundle lock --add-platform x86_64-linux
```
mzagaja, line 7: If your using ARM shouldn't this be `aarch64-linux`? It's a [supported platform for nokogiri](https://nokogiri.org/tutorials/installing_nokogiri.html#supported-platforms).  

That said `x86_64-linux` is [already available in the current Gemfile.lock](https://github.com/codeforboston/urban-league-heat-pump-accelerator/blob/main/backend/Gemfile.lock#LL241) so you shouldn't have to run this every time.

zganger:
I used this solution from StackOverflow- for arm64 it does not automatically build but must be included manually and requires additional glibc installation:

https://stackoverflow.com/questions/70963924/unable-to-load-nokogiri-in-docker-container-on-m1-mac

The errors disappeared after applying the fix.

mzagaja:
Sounds good, if it works, it works! Thanks for getting this pulled together 😄 .

PR #118, opened 2022-12-13
Adds confirmation modal to delete button on admin user profile

caseydulong:


Zevez7:
The security checks is from rails and not on our part. We can still pull and push for right now and let backend deal with the security check.

PR #126, opened 2023-01-14
Removed the previous design from the public view home and added the begin of the new design.

thiagobardini:
Added a theme file for standard style.
Added react-router-dom.
Created a File Structure.
Added layout folder for Footer and Navbar components.
Removed the unnecessaries files in the public folder.
Added standard font to the project - Roboto.
Created ButtomGetPump component 
Added content to the Home Page
Created reusable Cards components


frontend/front/src/pages/Public/Components/CardBenefitsSection.js
```diff
@@ -0,0 +1,107 @@
...
+  display: "flex",
+  flexDirection: "column",
+  justifyContent: "center",
+  textAlign: "center",
+}));
+
+const CardBenefitsSection = ({ cards }) => {
+  return (
+    <>
+      <Typography gutterBottom variant="h4" mb={6} align="center" gutterBottom>
```
Zevez7, line 23: extra gutterBottom that should be deleted

frontend/front/src/pages/Public/Components/CardHero.js
```diff
@@ -0,0 +1,55 @@
+import React from "react";
+import { Typography, Box, Grid, CardContent, CardActions } from "@mui/material";
+import ButtonGetPumnp from "./ButtonGetPump";
+
+const CardHero = ({ title, paragraphs, image }) => {
+  return (
+    <Grid
```
Zevez7, line 7: So, I think the Grid here flex alignment is messing up the alignment when it's in the mobile version. There's a fix to this, but I don't remember. It has something to do with spacing it around or evenly. I'm going to left you find out how to fix it.

frontend/front/src/pages/Public/Pages/About.js
```diff
@@ -0,0 +1,39 @@
+import React from "react";
+import { Box, Grid, Typography } from "@mui/material";
+
+const About = () => {
```
Zevez7, line 4: I do not agree with how you're creating your page folder structure here. Putting page component into the Components folder is ok for small project. But once you start having many pages with lots of component, you Components folder will get to large with multiple components from many pages. It's better to make sperate folders in Pages folder for all the pages and add the component that is stored in that pages for that page folder. The Components folder can be a store for Global components that is placed into multiple pages. I think it's better to make this changes early on before later when components folder gets too big and too confusing.

frontend/front/src/pages/Public/Assets/theme.js
```diff
@@ -0,0 +1,59 @@
+import { createTheme } from "@mui/material";
```
Zevez7, line 1: I don't think a theme right now is necessary and it might complicate the front end development for other developers. If you are comfortable with it and can show other developers how to use it then I'm ok with it.

Zevez7 (approved):
Aside from style change. The design is overall good. I can approval of this for merge and we can talk a little more about design choices during the review on Tuesday. Good work. Glad to see you take the reign and make it your own.

Zevez7:
Will approve now for merging. Will need a update to some changes later on.

frontend/front/src/pages/Public/Assets/theme.js
thiagobardini: I am still learning how to use the createTheme function. This is a new approach in the latest version of Material-UI, utilizing variants for each MUI component (Paper, Button, Input...) . In my opinion, this could be an effective way to standardize the styles throughout the application. I will continue to learn and will provide an update within the next couple of weeks on whether or not it is a viable alternative.

frontend/front/src/pages/Public/Pages/About.js
thiagobardini: I am open to changing the folder structure. When I first set it up, I was uncertain if it was the most efficient approach. Let's discuss it tonight so we can implement the necessary changes to the folder structure.

frontend/front/src/pages/Public/Components/CardHero.js
thiagobardini: I noticed it is a little messy. I will work on fixing it this week and create the mobile version.

thiagobardini:
Alright! I will merged it and I will review your messages to address the issues. However, I think it is important that we make the new design available to everyone so that others can assist us in improving it.

PR #131, opened 2023-01-18
Remove user association from survey response

mikelynch:
We already have an association to user from `survey_visit`, so do not need another association from `survey_response` which could be inconsistent. If we need to get back to the surveyor from a `survey_response`, then we can join through the visit type.


mikelynch:
@mzagaja , do you see the cause here? When I ran `db:migrate` in the CodeSpace it ran fine, and updated `schema.rb` the way I'd expect but it's failing in the CI.

mzagaja:
I can take a look tonight. The error makes it sound like we need to validate the schema against our declarations in the models for any mismatches. It could also be an older migration is out of sync.

mikelynch:
I think I've worked out the issue, though I'm not sure how we should fix it – just putting something on the Slack channel about it.

mikelynch:
I think this would probably be fixed having merged in PR #133 but there seem to be some newly reported security vulnerabilities with the current pinned packages. I'd like confirmation that this is all working before we merge this one in, so I've just made a separate PR to sort the packages – see #134 .

mikelynch:
As we'd hoped, fixing the migration order means this new migration now succeeds – I think this is ready to merge.

mzagaja (approved):
🚢 it!

PR #134, opened 2023-01-20
Update package dependencies

mikelynch:
I was working on a different PR, which I think is probably okay now, but on the CI run it fails the security checks which is unfortunately one of the first steps so I'm not sure. It looks like in the last few days there must have been some newly disclosed package vulnerabilities. I just ran `bundle update` here to sort these – can we see any risks from doing that?


zganger:
From a review of the changes, most of the updated packages are only a hotfix version increment (ie. Semantic versioning sub-patch increment, x.x.x -> x.x.x.1). I don't think these changes will affect the app.

mikelynch:
Agreed @zganger , most of these were from the Rails 7.0.4 -> 7.0.4.1 security release. There are a few semantic versioning minor version changes as well, but those shouldn't introduce breaking changes.

zganger (approved):

PR #135, opened 2023-01-20
Add data model docs

mikelynch:
With recently joining the project, I've been learning about how things work. In my experience, having more documentation (as long as it's reasonably up to date!) is better than less, so I've written up some of my notes. Would these be useful for other contributors to the project?


lhuhmann:
I was thinking about suggesting this myself - I think it's a great idea. I agree that the main concern I had was whether the team has the capacity to keep documentation up-to-date. I was hoping to hear from some folks who have been around Code for Boston longer to learn more about their experience with documentation in that context. Personally, I definitely find documentation like this useful when getting up to speed on a project.

mzagaja:
This is great. Thanks Mike.

PR #140, opened 2023-01-26
Change some string property assessment columns to booleans

### LINKED ISSUE #45: Migrate Columns from Strings to Booleans
In the data schema `:own_occ` and `:corner_unit` need to have migrations written to convert them from String types to Booleans or we need to figure out whether we shoudl use enum or keep them String types depending on the data we want to preserve. Needs further investigation as to what values are in the field, get some examples of the outliers and the we can ask Christopher Scranton.

mikelynch:
I made a basic PR to resolve #45, which just updates the types for these columns.

I don't think it will matter (as we'll upload the data starting from the latest schema), but just in case – I'm not sure what this would do if the data had already been loaded with the "Y"/"N" values that @lhuhmann previously identified from the CSV files. I did try doing something like the following, but I'm not sure if this is allowed in Rails with the PostgreSQL adapter as I get some column `PG::UndefinedColumn` errors:
```ruby
# up
change_column :property_assessments, :own_occ, :boolean, using: "UPPER(own_acc) LIKE 'Y%'"

# down
change_column :property_assessments, :own_occ, :string, using: "CASE WHEN own_acc THEN 'Y' ELSE 'N' END"
```
Hopefully this doesn't matter, as we'll just load the data fresh but I wanted to flag it.


mzagaja:
@mikelynch You might be able to use `ActiveRecord::Base.connection.execute(SQL_GOES_HERE)` to execute raw SQL as an argument to `using` but I'd have to check/try that first.

mikelynch:
Does it matter if existing pre-loaded rows are not handled well by this migration given we don't have an existing deployed system quite yet? We can just make sure the script for this properly handles maps the string `"Y"`/`"N"` values to `true`/`false`. Assuming this is the case, I think this migration is ready to go.

mzagaja (approved):
Yeah we don't have to worry about doing the conversions, so you can ship this.

PR #163, opened 2023-02-09
Add front-end ReCaptcha support

### LINKED ISSUE #148: reCAPTCHA for public survey submission
https://developers.google.com/recaptcha/intro

Front end implementation looks pretty straightforward. We're going to need to get an API key from Google, so we should probably coordinate with the back-end team on that since we don't want to use someone's personal account for that.

mattdelsordo:
Follow-up required:
1. We need an API token https://www.google.com/recaptcha/admin/create
2. (1.5): Add that API token to the build process so that it gets injected into the site on deployment (add add it to local .env files)
3. the ReCAPTCHA API returns a token that we need to pass to the back-end, the back-end then handles the request to validate that token with Google. It'll get back a score that'll tell us what % likelihood the submitter was human and then we can do whatever we want with that info https://developers.google.com/recaptcha/docs/verify

Resolves #148


Zevez7:
Ok. I have reviewed this.

PR #173, opened 2023-02-20
Merge surveyor survey with public survey for reusable component

mattdelsordo:
I recreated the surveyor survey to reuse the public survey code. Questions/options are based on a google doc I found in the backend slack channel and probably not final, but here's something.


Zevez7 (approved):
good work,

Zevez7:
There's a code conflict that needs to be resolved before it can be merged. Please check this out.

Zevez7:
The apt number should be optional on the public survey. It's currently required.

Zevez7:
The jsx file should be changed to js file to keep the all the file in the same file type. I don't think it has any conflict with js, but to keep it uniform it would be good.

PR #184, opened 2023-02-23
Add display order attribute to survey_question type

mikelynch:
So far this is very simple – just adding a single attribute to survey question, so that we can specify the display order. 

Questions:
* Do we want to enforce a uniqueness constraint on `(survey_id, display_order)` so that it is impossible to have two questions in a survey with the same display position?
* Rubocop complains if you have a non-null column without a default value. Think we should just e.g. make that 0 here?
* I think we probably want to return the survey_questions (in order) when we call `GET /surveys/<id>` – the survey object itself is very uninteresting otherwise. @mzagaja , think it makes sense to add that logic to the surveys controller in this PR?


aTegart:
@mikelynch 

I'll also let @mzagaja chime in, but here's my thoughts:

> * Do we want to enforce a uniqueness constraint on `(survey_id, display_order)` so that it is impossible to have two questions in a survey with the same display position?

I think this comes down to how often we expect the survey to be edited via adding/removing/reordering questions (vs just making a new one). Inserting a question in the middle of the survey would require updating the `display_order` values for every question that came after it. But I think such a change is unlikely, and we'd probably just make a fresh 'survey version 2' with the update. In summary, having a determininstic order seems good and we're unlikely to need to update it, so uniqueness sounds good to me.
> * Rubocop complains if you have a non-null column without a default value. Think we should just e.g. make that 0 here?

If we both default to 0 for new questions and have a uniqueness constraint, and if we also start our question ordering at 0, then hopefully that constraint will prevent us from adding new questions without explicitly assigning them a non-default `display_order`. So I think a default of 0 makes sense. (I was considering suggesting defaulting to max int instead, but there doesn't seem to be a concept of max int in ruby https://stackoverflow.com/questions/535721/ruby-max-integer)
> * I think we probably want to return the survey_questions (in order) when we call `GET /surveys/<id>` – the survey object itself is very uninteresting otherwise. @mzagaja , think it makes sense to add that logic to the surveys controller in this PR?

I'll leave this one to @mzagaja given his experience thinking about surveys, but this makes sense to me. If you do want to define your own ordering for the questions as according to `display_order`, perhaps this will help: https://apidock.com/rails/ActiveRecord/QueryMethods/order

mikelynch:
Thanks @aTegart, that makes sense to me. I think I've added uniqueness constraints on (survey_id, display_order), and set the default value 0 (which sorts the rubocop violations).

I've also added a `has_many` association to the survey model. I don't think it needs a migration to handle this as the other half is already present. I haven't quite figured out what we do to load the “nested” questions within a survey – will try to figure that out!

mzagaja:
This all seems reasonable to me. For the default value I wonder if we can just calculate it based on the existing value(s)? Maybe something like SurveyQuestion.where(survey_id: value).maximum(:display_order) + 1`

mikelynch:
I've just made a minimal version of this – now if you request /surveys.json or /surveys/1.json, you get back a lump of JSON including the survey_questions, sorted by `display_order`. I'd prefer this if it reused the jbuilder pieces from survey_questions, but didn't quite see the syntax to have that work properly. In any case, I'm not sure it makes a lot of sense to have `survey_questions` as top-level resources – we never want to get an individual one of them, it's always looking at them as part of a larger survey.

mikelynch:
There are some RSpec issues here that I need to double-check – or someone else is welcome to jump in!

mikelynch:
I think this is ready to go now

mzagaja (approved):
I tested and this should be good to go!

PR #227, opened 2023-03-15
Add dummy JSON data for front-end team to use in development

mikelynch:
Resolves #214

This adds dummy JSON for the surveyor, assignment and homes endpoints. The dummy JSON for the surveys is in the separate PR.

PR #230, opened 2023-03-15
Add new attribute to home to record order to visit within an assignment

### LINKED ISSUE #225: Add new attribute to homes to record order to visit within an assignment
At present, we have an `assignment_id` attribute to record which assignment a house belongs to. We also need to record the order to visit these in (possibly with a uniqueness constraint on (`assignment_id`, `assignment_visit_order`).

Requires changes to model, and the seed data.

mikelynch:
Resolves #225

This adds a new `visit_order` attribute to homes, which specifies the order to visit the homes within an assignment. (assignment_id, visit_order) should be unique (unless these are NULL, which is allowed). It also ensures that home has a visit_order if and only if it has an assignment_id (as these are only meaningful together).

It updates the home test data, to set this for homes.

This will require a small update to the exemplar JSON, but it's probably neater to do that after [PR 227](https://github.com/codeforboston/urban-league-heat-pump-accelerator/pull/227) is merged in.


mzagaja (approved):
This looks good to me.

PR #249, opened 2023-04-04
Giant survey related pull request

mattdelsordo:
This covers a bunch of stuff:

1. Created a rudimentary address validation page for the public view. It just does a POST to the normal `home` endpoint right now.
2. Added an address view to the survey component so its clear what address the survey submission is associated with
3. Success pages for both public and surveyor surveys
4. Made it so you can specify which survey to use in the survey component and added metadata questions specific to the surveyor survey
5. Wired the public + surveyor + admin survey pages up to jsonserver


frontend/front/src/dummyData/jsonServerData/homesurvey.js
mattdelsordo, line 1: I don't think the back end has example data for this so I guessed here

frontend/front/src/pages/Public/Components/ThanksForSubmission.js
```diff
@@ -0,0 +1,25 @@
...
+  const navigate = useNavigate();
+  return (
+    <Alert
+      severity="success"
+      action={
+        <Button
+          color="inherit"
+          size="small"
+          onClick={() => {
+            navigate("/public");
```
mattdelsordo, line 15: Not 100% sure how navigating to `/public` for example is going to interact with the deployed website

Zevez7 (approved):
ok it works good.

Issue #246, opened 2023-03-30
Public View Improvements


thiagobardini:
- Modified the background color of the footer and navbar, and removed several colors from the color palette.
- Improved the design of the footer and navbar and added new links

PR #250, opened 2023-04-04
Frontend 246 public view - Redesigned the Footer and Navbar to improve the overall user experience

thiagobardini:
- Modified the background color of the footer and navbar, and removed several colors from the color palette.
- Improved the design of the footer and navbar and added new links

Ref: https://github.com/codeforboston/urban-league-heat-pump-accelerator/issues/246


thiagobardini:
@Zevez7 FYI: These updates are related to the new footer and navbar design.

PR #261, opened 2023-04-05
API Endpoint with Vercel - dummy data

thiagobardini:
DO NOT MERGE IT YET. 

Run a local server for your JSON data and use it alongside Vercel. Here are the changes that have been made:

- Updated apiSlice.js, surveyorViewApiSlice.js, and adminViewApiSlice.js to automatically switch between local and Vercel-hosted APIs. This will ensure that you can use your JSON data when running the app locally, and switch to the Vercel-hosted APIs when deploying your app to production.
- Added a new script called "copy-db" to copy the contents of ./src/dummyData/jsonServerData/db.js to ./api/db.js. This will ensure that the database file is copied to the right location before the app is started.
- Modified the "start" script to run "copy-db" before starting the app using react-scripts. This will ensure that the db.js file is copied to the right location before the app is started.

Please review and merge this pull request if it meets your requirements.

Refs: 
- https://vercel.com/docs/serverless-functions/introduction 
- https://www.youtube.com/watch?v=xPLrHMGYpq4


frontend/front/src/redux/adminViewApiSlice.js
```diff
@@ -1,12 +1,36 @@
...
+//
+// To access and visualize the data from the Vercel-hosted API, simply
+// visit the API endpoint URLs using the Vercel deployment domain
+// (e.g., https://https://bhpa.vercel.app/api/your-endpoint).
+async function customBaseQuery(args, api, extraOptions) {
+  const localApiUrl = "http://localhost:3001";
+  const vercelApiUrl = "https://bhpa.vercel.app/api";
+
+  // Check if the local API is available
+  const isLocalApiAvailable = await fetch(localApiUrl)
```
mattdelsordo, line 15: Checking for the API like this seems kind of janky - because this is built into the baseQuery it's doing an extra request on EVERY request this api slice makes. IMO the better way to do this would be with an environment variable. Maybe Vercel has some sort of variable it makes available during the build process that you could use in the logic here to tell if its Vercel or not? Or you could go the other direction and use `process.env.NODE_ENV === "development"` to test if it was started with `yarn start`.

thiagobardini: Hi @mattdelsordo! Thanks for your feedback on the code. I agree that using an environment variable would be a better approach to checking if the API is available. I will look into the documentation for variables on Vercel and implement this change. Thanks again for your suggestion!

frontend/front/package.json
```diff
@@ -16,6 +16,7 @@
     "animate.css": "^4.1.1",
     "classnames": "^2.3.2",
     "clsx": "^1.2.1",
+    "dotenv": "^16.0.3",
```
mattdelsordo, line 19: Is dotenv actually getting used anywhere here? Regardless I think it's good to have.

mattdelsordo (approved):

PR #264, opened 2023-04-11
Form transitions + caching + bugfixes

mattdelsordo:
* added visual transitions between steps of each form
* fixed survey data not displaying properly after editing in admin view
* cache incomplete surveys in local storage until theyre submitted
* refresh list of survey visits with tag invalidation instead of reloading the page :sweat_smile:

### DIFF
--- a/frontend/front/src/components/SurveyComponent/SurveyComponent.js
+++ b/frontend/front/src/components/SurveyComponent/SurveyComponent.js
@@ -1,4 +1,10 @@
-import React, { useCallback, useEffect, useMemo, useState } from "react";
+import React, {
+  useCallback,
+  useEffect,
+  useMemo,
+  useState,
+  forwardRef,
+} from "react";
 import { useForm } from "react-hook-form";
 import { Button, Stack, Alert, CircularProgress, Box } from "@mui/material";
 import { HeatPumpDropdown } from "./HeatPumpDropdown";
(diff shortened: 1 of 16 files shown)


frontend/front/src/components/SurveyComponent/SurveyComponent.js
```diff
@@ -11,217 +17,230 @@ import { useNavigate } from "react-router-dom";
...
-  isLoading,
-  activeHome,
-  isEditable,
-  surveyId,
-  formSpacing,
-  defaultData,
-  onDelete,
-}) => {
-  const navigate = useNavigate();
+export const SurveyComponent = forwardRef(
```
mattdelsordo, line 20: Wrapping this in forwardRef caused the diff to be ridiculous - most of the code is the same here

frontend/front/src/pages/Admin/survey/SurveyProfile.js
```diff
@@ -60,9 +60,12 @@ const SurveyProfile = () => {
     navigate(0); // reload page to update data
   }, [deleteSurveyVisit, surveyVisitId, navigate]);
 
-  const PageContent = useCallback(() => {
```
mattdelsordo, line 63: Note to self: treating a useCallback like a proper react component causes rendering issues

Zevez7 (approved):
tested the code and it looks good.

PR #271, opened 2023-04-12
228 two surveyors per assignment

mikelynch:
Creating a draft PR here to put comments on…


mikelynch:
I was thinking a bit more about how to represent the home–assignment association, because I wonder whether we're painting ourselves into a corner right now. Assumptions:
* We fundamentally want an assignment to represent a group of homes – i.e. there is a 1:many relationship between assignment and homes. Fundamentally, this is a property of the _assignment_, not the _home_ (i.e. the home means something independently of the assignment – but the assignment without homes isn't meaningful).
* The new join table feels like a good way to represent the many:many relationship between surveyors and assignments.

Given that, I can see a few different ways of capturing the 1:many assignment to homes relationship:
1. Represent this the way we started – i.e. there is one row in `assignments` for each assignment, and one row in `homes` for each home. The assignment that the home is in is represented by a `assignment_id` column on `homes`, which is either `NULL` (i.e. unassigned), or indicates the ID for the assignment. To find which homes are in an assignment, we simply join on `assignment.id = homes.assignment_id`. This allows efficiently querying either way – we can look up the assignment of a home, and also go from an assignment to all homes in it. This requires no changes to the many:many surveyors to assignments relationship. 
image
2. Keep a single row per home; keep a single row per assignment, but have an array of the homes in it – i.e. the `homes` table does not have an `assignment_id` column. Instead, the `assignments` table has a `home_ids` array column, which consists of the homes in that assignment. We can easily go from an assignment to the list of homes, but now, to look up which assignment a home is in, we have to do a scan through the `assignments` table, checking each array for that `home_id`. In the event that this home is unassigned, we would have to go through the entire table before finding it is not in any. This also allows representing the same home in more than one assignment (unless a unique constraint on contents of arrays is possible?), which we don't want. This also requires no changes to the many:many surveyors to assignments relationship.
image
3. This is what we've started doing on the branch – keep a single row per home, but now have **multiple rows per assignment** – with `assignment.cluster_code` indicating the logical assignment (i.e. all `assignments` rows with the same `cluster_code` are one logical assignment). There should be at most one entry in `assignments` with the same `home_id` (this could be neatly handled with a unique constraint). This does support querying in both directions efficiently – we can go from a home, and search `assignments` for that home_id. We can also start from an assignment `cluster_code`, and find all entries with that value. However, now `assignment.id` does not have any real meaning, so we would need to update the many:many relationship from surveyors to assignments to use `cluster_code` instead. This also has the unconventional property that there are many rows for a logical entity, which isn't true for option 1.
image

@mzagaja , I know you thought option 3 might make it easier to insert the assignments in the first place, but it does feel a bit less like the “normal” way of representing this 1:many relation. I'd lean to keeping this bit as it was before (i.e. option 1). See below for what might be involved:

## Inserting initial data
Do we know how we're getting the assignments info from the data team? My understanding is that they don't have names – they're simply sets of sets of homes.

If the homes already have a unique numeric ID, we could just use that in the DB. In that case, the assignments would just be a set of sets of those IDs (e.g. `{ {1, 2, 3}, {4, 5, 6, 7} }` would indicate two assignments, one with 1–3, and the other with 4–7). The options would then require:
1. Give each assignment an ID (we can just run through them in sequence). Create a row for each. Run through all the homes records, and use the assignment ID we created before.
2. Give each assignment an ID (we can just run through them in sequence). Create a row for each, and include the set of home_ids in the record.
3. Give each assignment an ID (this is the `cluster_code`). For each home in each assignment, create an assignment record, using the home ID, and referring to this cluster_code.

If the homes don't already have unique IDs (this seems unlikely?), then presumably the assignments look more like { {1 Main Street Unit, …}, … }. First we have to create all the homes, and give them arbitrary home IDs – but then we do the same as before.

In either case, I feel like the slightly simpler steps here for option 3 doesn't really justify the unusual way of representing it.

What does everyone else think? Have I missed something here – maybe another factor to consider?

lhuhmann:
I'm with you on preferring option 1. I think it is easier to understand and explain to other people (new backend team members or members of other teams the backend team interfaces with) what's going on conceptually in that scenario, since the database more directly reflects how the overall team conceptualizes of things. If I'm remembering correctly, there are "clusters" and "regions" (groups of clusters) coming from the data team. I don't think the data team or the front end team have an additional conceptual need for an "assignment" in the sense that it is used in scenario 3, and I'm not convinced that we (the backend team) really benefit from adding such a concept either. I agree that it seems like the fundamental relationship we're trying to capture here is a 1:many relationship between assignments and homes, and that scenario 1 seems like a more standard way of representing that.

mikelynch:
Just to confirm what we discussed yesterday – we thought option 1 made most sense. I did an interactive rebase to remove the commit that changed that part, and have force-pushed back on this branch (sorry!) For reference, the old branch still exists at [228-two-surveyors-per-assignment-orig](https://github.com/codeforboston/urban-league-heat-pump-accelerator/tree/228-two-surveyors-per-assignment-orig). I'll flag when this is ready to look at again.

aTegart:
Since it looks like the security checks that are failing are the ones you addressed in #275, can we just merge main in here and expect things to probably work?

PR #284, opened 2023-04-26
Added (and fixed) filtering for surveyors and homes

### LINKED ISSUE #179: Add filtering on assignment_id when getting homes
Support filtering homes which do not have an assignment (i.e. `home.assignment_id == null`)

`HTTP GET /homes/` with some filter?

aTegart:
Resolves #179


backend/app/controllers/surveyors_controller.rb
```diff
@@ -6,7 +6,7 @@ class SurveyorsController < ApplicationController
 
   # GET /surveyors or /surveyors.json
   def index
-    @surveyors = Surveyor.where(params.slice(:city, :zipcode, :state, :role, :status))
+    @surveyors = Surveyor.where(search_params.slice(:city, :zipcode, :state, :role, :status))
```
mzagaja, line 9: I don't think you need to use `.slice` here since we're already using `permit` to allow these.
```suggestion
    @surveyors = Surveyor.where(search_params)
```

backend/app/controllers/homes_controller.rb
```diff
@@ -5,7 +5,8 @@ class HomesController < ApplicationController
 
   # GET /homes or /homes.json
   def index
-    @homes = Home.all.first(30)
+    @homes = Home.where(search_params.slice(:street_number, :street_name, :unit_number, :city, :state, :zip_code,
+                                            :building_type, :assignment_id))
```
mzagaja, line 9: Think you can kill the `.slice` here and also have this work.

```suggestion
    @homes = Home.where(search_params)
```

mzagaja (approved):
Yeah these look good to me. The security issue is just to complain about being able to make some consequential updates to the `User` using the regular controllers, it's often the case that something like ActiveAdmin is used to change those and you don't want to expose them via the normal routes, but if we're doing an API with a front end, that's how we need to do it. So just a heads up type warning.

mikelynch:
Thanks @mzagaja ! I was reasonably optimistic it was okay but didn't want to just blanket-ignore flagged security issues…

Issue #252, opened 2023-04-05
Add back-end endpoints to rtk-query setup, mocked with jsonserver


mattdelsordo:
Regardless of how they get used, we need to connect the back-end endpoints up to RTK query. Part of this is already done but we're missing a bunch. Having all this done up front will save us time when we have to hook it up to whatever components.

If you're familiar with Rails you can just look at the back-end code in `/backend` for reference. The API controllers live in `/backend/app/controllers`. Here are some other docs that might help:
- backend API reference: https://docs.google.com/document/d/1QAek7K-pvnCJe4wxm38QdPCbszEE-esw3w-tuhEtMDQ/edit#heading=h.l6tzh4whd8cj
- rtk query docs: https://redux-toolkit.js.org/rtk-query/overview

For endpoints we currently have:

- `homes`
- `user` (should actually be `surveyors`)
- `surveys`
- `homesurvey` (should be `survey_visits`)

and we need:

- `assignments`
- `property_assessments`
- `survey_answers`*
- `survey_responses`*

*to be honest I'm not sure what the utility of these guys is, I kind of thought these would be covered by survey_visits

A wrinkle is that we have three instances of `createApi` for some reason. Per the RTK Query docs: 

> In most cases, you should use this once per app, with "one API slice per base URL" as a rule of thumb.

so we should probably consolidate `src/api/apiSlice.js`, `src/redux/adminViewApiSlice.js`, and `src/redux/apiSlice.js` into a single file.

Additionally, you're going to have to add new "tables" to the `jsonserver` config in `frontend/front/src/dummyData/jsonServerData`. Each rails controller needs an equivalent `.js` file here that is referenced in `db.js`. `/backend/db/schema.rb` defines the shape of the data as it is in the database. For now don't worry about getting fancy with json-server, just focus on the CRUD endpoints.

mattdelsordo:
Done

PR #289, opened 2023-05-01
apiSlice consolidation and expansion

LWest001:
This PR is in response to **issue #252 : Add back-end endpoints to rtk-query setup, mocked with jsonserver**

**Here's what's been done:**
✅Merge `src/redux/apiSlice` into `src/api/apiSlice.js`
✅Generate CRUD endpoints for:
- `assignments`
- `property_assessments` *
- `survey_answers` *
- `survey_responses` *

✅Normalize names of slice methods (start with get, update, create or delete/end with Data)
✅Add tagTypes:
- `Surveyor`
- `Survey`
- `Assignment`
- `PropertyAssessment`

✅Point all imports to `/src/api/apiSlice.js` with the correct hook names
✅Create jsonserver config files for (*) above, along with ones for `survey_visits`, `survey_questions` and `surveyors`.
✅Add new jsonserver config files to db.js

❌The jsonserver config files do not have data included - they are just templated with null values. I included comments for each key with the datatype and any other properties given in the schema (eg. `null: false`).

**Other issues that came up along the way:**
- I'm not 100% sure I set up the caching tags correctly (ie `providesTags` and `invalidatesTags` properties within endpoint methods)
- There are two other files using `createApi()` within `src/redux` called `adminViewApiSlice.js` and `surveyorViewApiSlice.js`. I'm a bit confused on these - the slice in `adminViewApiSlice.js` is itself called `surveyorViewApiSlice`, which is the same name as the slice inside `surveyorViewApiSlice.js`. Both have different endpoints, some of which are the same as endpoints within the main `/api/apiSlice.js`. I left these in place for now as I'm not if they are deprecated or still in use, and which should be saved vs merged. Notably, `surveyorViewApiSlice.js` uses a different base URL than all the other api slices.
- Most DELETE operations are crossed out on the Google doc. I left them in place, but maybe they should not be included in the apiSlice.
- The word "Data" in the name of all apiSlice methods is probably not necessary


frontend/front/.yarnrc.yml
mattdelsordo, line 1: Did you add this on purpose? I don't think it ought to be necessary. Is this why `.yarn` was generated? In either case we definitely don't want `.yarn/cache` checked in to version control, please add that to the `.gitignore` so we don't have to store all these zips in here.

frontend/front/src/api/apiSlice.js
```diff
@@ -1,83 +1,328 @@
...
-      invalidatesTags: ["User"],
+      invalidatesTags: ["Survey"],
+    }),
+    // mocking surveyorView data
+    // getAssigment within surveyor view
+    getSurveyorAssigment: builder.query({
+      query: () => ({
+        url: `/surveyor/assignment/user1`,
+        method: "get",
+        mock: SurveyorViewAssigment1,
```
mattdelsordo, line 124: Remove this mock, we shouldn't need this with json-server

frontend/front/src/api/apiSlice.js
```diff
@@ -1,83 +1,328 @@
...
+    getAssignmentsData: builder.query({
+      query: () => "/assignments",
+      transformResponse: (res) => (res ? res.sort(sortById) : []),
+      providesTags: [{ type: "Assignment" }],
+    }),
+    getAssignmentData: builder.query({
+      query: (id) => `/assignments/${id}`,
+    }),
+    createAssignmentData: builder.mutation({
+      query: (Assignment) => ({
```
mattdelsordo, line 243:
```suggestion
      query: (assignment) => ({
```

frontend/front/src/dummyData/jsonServerData/property_assessments.js
```diff
@@ -0,0 +1,70 @@
+module.exports = [
+  {
+    created_at: null, // type: datetime, null: false
```
mattdelsordo, line 3: Good call with these type comments IMO. Make sure all these new guys have an `id` just for json-server's sake

frontend/front/yarn.lock
mattdelsordo, line 1: I'm assuming this changed because of the `.yarnrc`?

mattdelsordo (changes requested):
Looks good! Please squash these changes before or when you merge though

mattdelsordo:
Re: your questions, I think you can merge the admin/surveyor view API slice into `apiSlice.js`. As for the DELETE endpoints, I say leave them, if we end up not needing them we wont use them but we might as well have all the CRUD operations for now. I agree that we don't need the redundant "data" in the hook names. As for the caching tags, [here's the advanced tutorial if you didn't find it](https://redux.js.org/tutorials/essentials/part-8-rtk-query-advanced#invalidating-specific-items).

PR #308, opened 2023-05-03
Frontend main

LWest001:
The RTK Query apiSlice is all set up. See: [PR #289 ](https://github.com/codeforboston/urban-league-heat-pump-accelerator/pull/289). 

The purpose of this pull request is to confirm that this should be merged into main as we migrate away from the Frontend_main branch.


thiagobardini:
I am still in the process of learning RTK Query, so I am not able to review your code. However, I tested your requests and was able to delete a home (/homes), but I was unable to create one (got a 404 error). Anyway, I look forward to learning more with you guys

mattdelsordo:
@thiagobardini I already reviewed this on the merge into `Frontend_main` so we should be good here

PR #309, opened 2023-05-09
Add geocodes to assignments model

### LINKED ISSUE #273: Add geocode to assignments model
We should record the geocode for the centroid of each assignment/cluster (as provided by data team) – stored as a string something like `42.366867902661276, -71.09082818340268`.

mikelynch:
Closes #273 

This adds a single column for geocode to the assignment model.


mzagaja (approved):
Yeah this seems good for a v1 for me. We can work on migrating to real geospatial columns later.

mikelynch:
I think this matches the way we're storing geocodes elsewhere anywhere, so hopefully all good! If necessary we can adjust both (all?) others later.

PR #310, opened 2023-05-09
Add filtering functionality to survey visit endpoint

### LINKED ISSUE #105: Potentially add filtering functionality to survey_visit endpoint
### Overview
We need a suvery_visit (previously HOMESURVEY) endpoint.

Attributes
- Id
- homeId
- surveyorId 
- surveyId
- surveyCompleted
- True, false

image

image

see https://www.forcedive.com/surveyor/dashboard

Left To do:
- [ ] Limit to 30 items
- [ ] Add filtering.

mikelynch:
Resolves #105

### DIFF
--- a/backend/app/controllers/survey_visits_controller.rb
+++ b/backend/app/controllers/survey_visits_controller.rb
@@ -5,7 +5,7 @@ class SurveyVisitsController < ApplicationController
 
   # GET /survey_visits or /survey_visits.json
   def index
-    @survey_visits = SurveyVisit.where(search_params).first(30)
+    @survey_visits = SurveyVisit.where(search_params)
   end
 
   # GET /survey_visits/1 or /survey_visits/1.json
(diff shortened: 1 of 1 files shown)


backend/app/controllers/survey_visits_controller.rb
```diff
@@ -5,7 +5,7 @@ class SurveyVisitsController < ApplicationController
 
   # GET /survey_visits or /survey_visits.json
   def index
-    @survey_visits = SurveyVisit.all
+    @survey_visits = SurveyVisit.where(search_params).first(30)
```
mzagaja, line 8: Unless we think a search is unlikely to return more than 30 results I think it might be worth using [Kaminar](https://github.com/kaminari/kaminari)i to paginate here.

mikelynch: Yeah, I wondered about this before – the google doc spec said take 30, but I agree this could be a problem. I think if we're filtering to a specific home that should be fine, but if we want to filter more generally there could be too many results.

mikelynch: Okay, have removed the `.first(30)` part – does this otherwise look good to go?

PR #318, opened 2023-05-15
201: Get front end authentication working

### LINKED ISSUE #201: Get front-end authentication working
The backend team is using [devise](https://github.com/heartcombo/devise) for authentication. We need to integrate this with the front end so surveyors and admins can log in properly.

I think since we have a separate single-page-app (instead of using Rails views) back-end will have to use [devise-jwt](https://github.com/waiting-for-dev/devise-jwt). There's an rtk-query tutorial for handling authentication on the front end: https://redux-toolkit.js.org/rtk-query/usage/examples#authentication

Refs:
- https://dev.to/olumidesamuel_/implementing-protected-route-and-authentication-in-react-js-3cl4
- https://medium.com/swlh/react-reactions-cfdde7f08dff

Screenshot 2023-03-09 at 11 56 48 AM

mattdelsordo:
Resolves #201 

Note: I made a bunch of changes to the back end code to get this working that I need to discuss with the team/shouldn't merge, but they're here as an example.


backend/config/initializers/cors.rb
mattdelsordo, line 1: I think we need something to handle CORS on the back end so I added `rack-cors` here, but these are like the most permissive settings possible and we should get more specific with what we need.

backend/app/controllers/application_controller.rb
```diff
@@ -2,4 +2,5 @@
 
 class ApplicationController < ActionController::Base
   # before_action :authenticate_user!
+  protect_from_forgery with: :null_session
```
mattdelsordo, line 5: We need to discuss whether we actually need CSRF protection, I don't think so since we're using JWTs but I'm no expert, and there are other security considerations there

frontend/front/public/index.html
```diff
@@ -1,35 +1,25 @@
...
+  <link rel="icon" href="%PUBLIC_URL%/iconHeatPump.svg" />
+  <meta name="viewport" content="width=device-width, initial-scale=1" />
+  <meta name="theme-color" content="#000000" />
+  <meta name="description" content="Web site created using create-react-app" />
+  <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto:300,400,500,700&display=swap" />
+  <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons" />
+  <link href="https://fonts.googleapis.com/css2?family=Lora&display=swap" rel="stylesheet" />
+  <title>Boston Heat Pump Accelerator</title>
+</head>
+<!-- <script
```
mattdelsordo, line 15: Just commented out this API because the `%REACT_APP_GOOGLE_API_KEY%` is invalid... do we need to get an API key here?

frontend/front/src/api/apiSlice.js
```diff
@@ -1,12 +1,23 @@
...
   reducerPath: "apiSlice",
   baseQuery: fetchBaseQuery({
-    // baseUrl: "http://localhost:3001",
-    baseUrl: "https://testing-ibhy.onrender.com",
+    baseUrl,
+    prepareHeaders: (headers, { getState }) => {
+      const token = getState().login.token;
+      // add auth to every request, if the token exists
+      if (token) {
+        headers.set(AUTHORIZATION_HEADER, token);
```
mattdelsordo, line 17: This is the important part, this adds the `authorization` header to every request, if we have it.

frontend/front/src/api/apiSlice.js
```diff
@@ -311,6 +322,22 @@ export const apiSlice = createApi({
         ...result.map(({ id }) => ({ type: "PropertyAssessment", id })),
       ],
     }),
+    //sessions section of the slice
+    loginUser: builder.mutation({
+      query: ({ email, password }) => ({
+        url: "/users/sign_in",
+        method: "POST",
+        // TODO: should this be hashed??
+        body: { user: { email, password } },
```
mattdelsordo, line 331: Do we need to like hash passwords in transit? Or is HTTPS sufficient?

frontend/front/src/dummyData/jsonServerData/authMiddleware.js
mattdelsordo, line 1: Whipped up this middleware to mock out the auth stuff in jsonserver! It doesn't currently handle the JWT stuff though.

frontend/front/src/features/login/Login.js
mattdelsordo, line 1: Moved this component into `features/login` since it now handles login for any user role. It'll redirect to the correct page on login.

frontend/front/src/features/login/loginSlice.js
```diff
@@ -1,31 +1,50 @@
...
-    setAuthenticated: (state, action) => {
-      state.authenticated = action.payload;
+    setLoginInfo: (state, action) => {
+      state.user = action.payload.user;
+      state.token = action.payload.token;
     },
   },
+  extraReducers: (builder) => {
+    // update state whenever user logs in successfully
+    builder.addMatcher(
```
mattdelsordo, line 20: This business automatically puts the `authorization` header in redux + localstorage on login

frontend/front/src/routing/ProtectedRoute.js
```diff
@@ -0,0 +1,16 @@
...
+import { useUserHasRoles } from "../features/login/loginUtils";
+
+/*
+ * Redirects user if they do not have the correct credentials
+ */
+export const ProtectedRoute = ({ children, allowedRoles = [] }) => {
+  const userIsAllowed = useUserHasRoles(allowedRoles);
+  if (!userIsAllowed) {
+    // user is not authenticated
+    return <Navigate to={routes.LOGIN_ROUTE} />;
```
mattdelsordo, line 13: If the user isn't logged in they'll be redirected to the login page if they try to go to an unauthorized page. If they're logged in but have the wrong role, they'll get redirected to the login page and then the login page will redirect to the approved dashboard page.

mattdelsordo:
Notes:

1. I moved the back end security junk into its own PR so this is mergeable
2. There is going to have to be future work to accommodate #322

frontend/front/src/api/apiSlice.js
mikelynch: This **shouldn't** be hashed – we need the plaintext password, and rely on the HTTPS/TLS connection to get it safe in transit.

mattdelsordo:
Reviewed in person

PR #338, opened 2023-05-18
Tweak some recaptcha properties

### LINKED ISSUE #266: Get API key for ReCAPTCHA
1. Login here with a Google account that we want to be associated with the project: https://www.google.com/recaptcha/admin/create
2. Label: whatever seems appropriate
3. `Score based (v3)` API key
4. Domain: our front-end URL
5. `ReCaptcha.js` in our front-end code is already looking for an environment variable called `REACT_APP_RECAPTCHA_KEY`, so we want to add that as an environment variable at build time in Vercel

mattdelsordo:
Something to note is that the front end will send a null token to the back end if anything goes wrong.

Once this is merged we can close #266

### DIFF
--- a/frontend/front/src/components/ReCaptcha.js
+++ b/frontend/front/src/components/ReCaptcha.js
@@ -1,4 +1,4 @@
-import { useEffect, useCallback } from "react";
+import { useCallback, useEffect } from "react";
 
 const RECAPTCHA_SCRIPT_ID = "RECAPTCHA_SCRIPT";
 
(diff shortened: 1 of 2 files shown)


frontend/front/src/components/ReCaptcha.js
```diff
@@ -33,8 +33,7 @@ export const useGetReCAPTCHAToken = ({ action }) => {
 
           resolve(recaptchaToken);
         } catch (err) {
-          // TODO: don't print the error once recaptcha actually works
-          console.error("ReCAPTCHA error", err);
+          console.error("Error obtaining ReCAPTCHA token!");
```
mattdelsordo, line 36: Don't leak this error the the console, just silently send a null value.

thiagobardini (approved):

Issue #304, opened 2023-05-03
Admin : Users: Clicking on a User, show actual data. Hook it up to rtk-query


Rando-cal:

LWest001:
Working on this - see leo-westebbe/issue-304-show-user-data

Issue #303, opened 2023-05-03
Admin: Users: table should display actual data


Rando-cal:
hook it up to rtk-query

PR #339, opened 2023-05-18
Leo westebbe/issue 304 show user data

LWest001:
In response to #303 and #304. UserTable now loads a list of Users (aka surveyors), and clicking on a user then loads that user.

I'm not sure if this is a good way to set up with react-hook-form. I set the `value` prop for each `<TextField/>`  to conditionally load its data from the RTK-query hook, or the initial value if null. If someone on our team is an expert with this library, please assign as a reviewer!


mattdelsordo:
Could you rebase this off main without the jsonserver changes? It'll make it easier to review.

LWest001:
> Could you rebase this off main without the jsonserver changes? It'll make it easier to review.

@mattdelsordo Done

mattdelsordo:
It still looks pretty janked up, would it be alright if I had a go at it?

mattdelsordo:
So I think some git nonsense has occurred and these changes got included in your other PR and are already on `main`, but in that case they've been reviewed and everything looks good, so we can probably just close this as well.

LWest001:
> So I think some git nonsense has occurred and these changes got included in your other PR and are already on main, but in that case they've been reviewed and everything looks good, so we can probably just close this as well.

You're right Matt. I'm going to close this issue.

PR #341, opened 2023-05-21
Support creating new homes when filling out survey

### LINKED ISSUE #190: Explore How to Add New Homes from Public Survey Submissions
As a user I might visit bostonhpa.org and want to submit a survey. As a part of this submission I will include my address. We need to figure out whether:

* We can simply import the entire Boston address DB and let the user select through filtering (somehow?) but this could be very messy in relation to the assignment groups because now we have a large chunk of un assigned homes and this will make using that interface a pain in the ass.
* We can potentially use Pelias or another geocoding API to validate a user String of their address and include it in the Survey submission. Then maybe the admin or Surveyor will go through and pick the appropriate Home based on this validated String and create the association. If we go this route we need to setup the validator, and also the front-end needs an interface for doing this.
* Third option is we YOLO let user enter a non-validated address and pray we can match it to the address list. I think this still will need a front end interface for connecting the home to the SurveyResponse after its submitted by a Surveyor.

mikelynch:
Closes #190

We wish to support anonymous internet users filling out the survey, rather than just having the surveys completed by our surveyors going door to door. However, this means that a user may be completing the survey for a home which is not in our database. We decided to add an extra page when these users can type in the details for the home, and we will afterwards attempt to canonicalize it (e.g. with something like Pelias). One of three outcomes will occur:
1. The canonicalized address does not exist in the database already. We update the form of the address and leave as is.
2. The canonicalized address already existed in the database. We point the survey visit at the original home record, and delete the new one.
3. We cannot canonicalize the address (i.e. Pelias does not recognize it). **TODO: decide what do we do with these, just delete them?**


mikelynch:
I've made a version of this with a fake implementation (that does not actually do anything!) but demonstrates doing the Active Job part.

It also wires up the initial seed part, but on second thoughts that might need revisiting as we should probably rate-limit our calls to whatever we use for this. Maybe something like [limiter](https://github.com/Shopify/limiter) could help here?

backend/app/controllers/homes_controller.rb
```diff
@@ -22,26 +22,42 @@ def edit; end
   # POST /homes or /homes.json
   def create
     @home = Home.new(home_params)
+    # By definition, a new home's address has not been canonicalized.
+    @home.canonicalized = false
 
     respond_to do |format|
       if @home.save
         format.json { render :show, status: :created, location: @home }
+        CanonicalizeAddressJob.perform_later @home.id
```
mzagaja, line 31: Conventionally this sort of thing is usually added as an [ActiveRecord callback](http://guides.rubyonrails.org/active_record_callbacks.html) in the relevant model.

backend/app/controllers/homes_controller.rb
```diff
@@ -22,26 +22,42 @@ def edit; end
...
+    unless home_params.keys.intersection(%w[street_number street_name unit_number city state
+                                            zip_code]).empty?
+      home_params[:canonicalized] = false
+    end
+
     respond_to do |format|
       if @home.update(home_params)
         format.json { render :show, status: :ok, location: @home }
+
+        CanonicalizeAddressJob.perform_later @home.id unless @home.canonicalized?
```
mzagaja, line 53: Also worth refactoring this functionality into model.

backend/app/controllers/homes_controller.rb
mikelynch: I think I updated this to do things with ActiveRecord callbacks – looks like it's behaving the way we want. Actually ends up being a bit neater which is nice

backend/app/controllers/homes_controller.rb
mikelynch: Same for here

mikelynch:
Might still need to think about rate-limiting:
> It also wires up the initial seed part, but on second thoughts that might need revisiting as we should probably rate-limit our calls to whatever we use for this. Maybe something like [limiter](https://github.com/Shopify/limiter) could help here?

mikelynch:
I've wired up the rest of this to unify the homes if it finds an existing match. We should still need to decide how to handle the lookup if Pelias fails to find the address – do we want to keep those records?

mikelynch:
This hopefully has everything ready *except the actual Peleas lookup*, and possibly rate limiting. If the lookup fails (i.e. Peleas is unable to determine an address), then we keep that record, but the home has a status of unrecognized (and will probably want filtering out by the data team when they perform any analysis).

If we get that implementation in place, hopefully this will be ready to go. If we pre-canonicalize all data inserted, hopefully the rate limiting isn't essential – the anonymous requests are unlikely to be very high frequency.

PR #343, opened 2023-05-22
fix recaptcha action not getting passed correctly

mattdelsordo:


LWest001 (commented):
I see that the hook no longer destructures the input (good). Why is the `RECAPTCHA_ACTION_PUBLIC_SURVEY` now an exported constant rather than passing the string to the hook directly from `SurveyPage`?

mattdelsordo:
Just thought it would be a good idea to not have it be a magic value in case it needs to be changed in multiple places in the future.

LWest001 (approved):
Looks good.

PR #344, opened 2023-05-22
move .prettierrc up a level

mattdelsordo:
This way it'll apply the same settings to both /front and /jsonserver


LWest001 (approved):
Looks _🌸pretty🌺_ good to me. From what I can tell you just moved the .prettierrc up and formatted the files!

PR #345, opened 2023-05-22
connect user table to rtk query

mattdelsordo:
@LWest001 I tried to unjank the git history here, let me know if I missed anything major.


mattdelsordo:
Okay no this is worse hold on a second.....

PR #346, opened 2023-05-22
remove .yarnrc.yaml and regenerate yarn.lock

mattdelsordo:
I _think_ `.yarnrc.yml` got added to `front/` and then I moved it into `jsonserver` by accident. I don't think we want `nodeLinker` set at all, afaik the default settings for yarn are faster.


LWest001 (approved):
Looks good to me.

LWest001:
I'm getting `yarn.lock` diffs running `yarn install` after pulling the most recent merges to `main`. Happening in both `frontend\front\yarn.lock` and `frontend\jsonserver\yarn.lock`. All packages seem to be the same but with slightly different settings or formatting. Not sure if this has to do with this PR or not, or if it matters. Here's an example of how 1 package looks in my VSCode diff:

image

PR #357, opened 2023-05-24
remove src/pages/Developer, add UI tests to git workflow

### LINKED ISSUE #352: Remove `src/pages/Developer`
The code in `src/pages/Developer` doesn't pertain to any actual features of this project and should be removed, or at least be removed from react router such that the content is inaccessible. Other code in the project _shouldn't_ reference anything in there, but if it does that'll need to be resolved as well.

mattdelsordo:
Resolves #352


frontend/front/src/pages/Developer/googlemap/GoogleMap.js
mattdelsordo, line 1: This file might be good for later reference if we want to add an embedded google map instead of redirecting.

thiagobardini (approved):

PR #366, opened 2023-05-31
Leo westebbe/issue 365 load surveyor survey

LWest001:
In response to issue #365.

Clicking an incomplete home loads Survey 1 and the data for the associated home.

I'm not clear if there should be a capability to load a different survey - there are 4 on the server, but only survey with `id: 1` has any data in the `survey_questions` array. Outside of this branch, the `SurveyorSurvey` component loads survey with `id: 2`, but that one is blank.

### DIFF
--- a/frontend/front/src/api/apiSlice.js
+++ b/frontend/front/src/api/apiSlice.js
@@ -108,7 +108,7 @@ export const apiSlice = createApi({
 
     /* Survey Endpoints */
     getSurveys: builder.query({
-      query: () => ({ url: "/surveys", method: "GET" }),
+      query: () => ({ url: "/surveys.json", method: "GET" }),
       transformResponse: (res) => res.sort(sortById),
       providesTags: (result = [], error, arg) => [
         "Survey",
--- a/frontend/front/src/api/apiSlice.js
+++ b/frontend/front/src/api/apiSlice.js
@@ -117,7 +117,7 @@ export const apiSlice = createApi({
     }),
     getSurveyStructure: builder.query({
       query: (id) => ({
-        url: `/surveys/${id}`,
+        url: `/surveys/${id}.json`,
         method: "GET",
       }),
       providesTags: (result, error, arg) => [{ type: "Survey", id: arg }],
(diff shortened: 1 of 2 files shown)


frontend/front/src/api/apiSlice.js
```diff
@@ -108,7 +108,7 @@ export const apiSlice = createApi({
 
     /* Survey Endpoints */
     getSurveys: builder.query({
-      query: () => ({ url: "/surveys", method: "GET" }),
+      query: () => ({ url: "/surveys.json", method: "GET" }),
```
mattdelsordo, line 111: I don't think adding `.json` was the correct fix since that's not an endpoint on the real back end, right? What was the problem here exactly?

LWest001: On `main`, clicking a home in a surveyor assignment encounters an error. I think this is because `/surveys` and `/surveys/:id` serve up HTML (see image)... Adding `.json` makes it work, but it's probably not the right fix.

So maybe this is an issue on the backend and not the frontend?

frontend/front/src/pages/Surveyor/Components/SurveyorSurvey.js
```diff
@@ -1,7 +1,7 @@
 import React, { forwardRef } from "react";
 import SurveyComponent from "../../../components/SurveyComponent/SurveyComponent";
 
-export const SURVEYOR_SURVEY_ID = "2";
+export const SURVEYOR_SURVEY_ID = "1";
```
LWest001, line 4: Here's what I'm referring to in the initial comment on this PR. Should this be a constant or is it meant to be dynamic? Why was it set to '2' before when only survey 1 contains any questions?

frontend/front/src/api/apiSlice.js
mattdelsordo: Oh weird! We should probably bring this up with the back end team if this is how the real server responds (if you haven't already)

frontend/front/src/pages/Surveyor/Components/SurveyorSurvey.js
mattdelsordo: Ah okay, I think this got missed somehow. The original idea was that there was going to be different versions of the survey for the public/surveyor use cases but we've moved away from that I guess, and as you point out there's only one survey in the actual DB. But it is meant to be hard-coded with the survey ID based on the use-case.

mattdelsordo (approved):

frontend/front/src/api/apiSlice.js
LWest001: I haven't. I'll squash and merge this then post on the backend Slack channel!

PR #379, opened 2023-06-05
Fix rubocop violations

mikelynch:
When the CORS stuff was added, it had some rubocop violations – this resolves them.

PR #380, opened 2023-06-05
Remove HTML templates for surveys endpoint

mikelynch:
@LWest001 found an issue where the `/surveys` endpoint defaults to HTML rather than JSON (it requires e.g. `/surveys/1.json` to force the JSON version, or including a content type header). This removes the HTML templates (which currently only exist for surveys anyway – the others were already removed).


aTegart (approved):
Do surveys already have their tests updated with the `as: :json` fixes (since html will no longer be available), or will that be needed after the rubocop fixes get merged in to this PR and stop blocking other tests from showing up? (Based on my skim of the test output, the rubocop test errors look the same as you've fixed in #379)

mikelynch:
@aTegart , I'd forgotten to do that – I rebased this branch off `main` with the rubocop fixes and have now adjusted those rspec tests. Hopefully all good now…

PR #382, opened 2023-06-05
Remove no-longer-need .json suffix from /surveys apiSlice endpoints

LWest001:
We no longer need to include the `'.json' suffix for these endpoints because @mikelynch updated the backend `/surveys` endpoints to deliver straight json (#380).


mattdelsordo (approved):
:clap:

PR #386, opened 2023-06-08
Enable PDF downloads on 'Get Involved' page

thiagobardini:
Hi, let me know if this is the best way to download these PDFs. I have created a folder in Urban League's Google Docs and I am using those links to open the PDFs.

Screen Shot 2023-06-08 at 4 08 35 PM


mattdelsordo (approved):
This looks fine/functional but I feel like the ideal way would be to host the documents at the website location itself instead of in Google Drive.

Upsides:
* documents are part of the website
* documents cant be removed or changed just whenever
* seems more secure to me? not sure if we want to be linking people to our Google Drive. we'll have to keep the sharing permissions up to date.


Downsides:
* harder to keep documents up to date
* more of an infrastructure question than a website code question

LWest001 (approved):
Works properly for me.

PR #387, opened 2023-06-09
HomeTable improvements

LWest001:
* Fix assignment column links to go to the correct URL
  * Previously went to a non-existent route
* Make this table behave like other tables
  * Previously, clicking the row navigated to the home profile 
  * Now there are explicit columns with links to assignment profile and home profile
  * No `onRowClick` function - this prevents admin from easily copying and pasting data if they wish and is inconsistent with other tables
* Remove AssignmentLink component - not needed now
* Make columns responsive/flexible width


frontend/front/src/pages/Admin/home/HomeTable.js
```diff
@@ -23,23 +22,64 @@ const getAddress = (params) => {
...
+      field: "zip_code",
+      headerName: "Zip Code",
+      minWidth: 100,
+      maxWidth: 150,
+      flex: 0.8,
+    },
+    {
+      field: "completed",
+      headerName: "Completed",
+      renderCell: (params) => (params.row.completed === "true" ? "Yes" : "No"),
```
mattdelsordo, line 48: I know `completed` is a matter of debate right now but this looks good as a baseline :)

mattdelsordo (approved):

PR #390, opened 2023-06-13
Leo westebbe/frontend 334 submit survey

### LINKED ISSUE #334: Wire survey up to rtk-query
See #325 for the related back end change.

Survey payload to the back end should look something like the following:
```
{
    "survey_visit": {
        "surveyor_id": 1,
        "home_id": 1,
        "survey_response_attributes": {
            "survey_id": 1,
            "completed": false,
            "survey_answers_attributes": {
                "1": {
                    "survey_question_id": 1,
                    "answer": "Yes"
                },
                "2": {
                    "survey_question_id": 2,
                    "answer": "No"
                }
            }
        }
    }
}
```
Update the jsonserver data accordingly because it is likely out of date. `surveyor_id` should be null in the public view case.

For ReCAPTCHA support, back end is working on that here: #327
I think they said they wanted the token and action to be passed in as headers, NOT in the body.

LWest001:
Closes #334. 

Here's what's working and what has changed:
- [x] Can submit a survey via `surveyor/house/{homeId}`
- [x] Can _almost_ submit a survey via `public/survey` - see below
- [x] `createSurveyVisit()` in `apiSlice` accepts a recaptcha token and attaches it as a header
- [x] `SurveyPage` gets a recaptcha token using the `create_survey` action
- [x] Error loading `SurveyComponent` is fixed (the `Loader` in `SurveyComponentWrapper` is now inside a div that can receive a ref)

Here's what needs to be ironed out:
- [ ] No access to who is conducting a survey, so `surveyor_id` is hard-coded in on submission of `survey_visit` in surveyor view
- [ ] Cannot submit a public survey because backend requires a real `surveyor_id` (cannot be null)
- [ ] Is the data being stored properly? `survey_visits` and child `survey_responses` submitted via these pages don't display all attributes when queried:
  - `survey_visits` don't return a `home_id` despite including one on submission
  - `survey_responses` don't return a `completed` attribute despite including one on submission
  - Neither `survey_responses` nor `survey_visits` return with a foreign key linking to each other - this would prevent us from being able to link a survey response and its answers to a survey visit, which I think is a necessary function?

### DIFF
--- a/frontend/front/src/pages/Public/Components/PublicSurvey.js
+++ b/frontend/front/src/pages/Public/Components/PublicSurvey.js
@@ -4,10 +4,12 @@ import SurveyComponent from "../../../components/SurveyComponent/SurveyComponent
 const PUBLIC_SURVEY_ID = "1";
 
 export const PublicSurvey = forwardRef((props, ref) => (
-  <SurveyComponent
-    {...props}
-    surveyId={PUBLIC_SURVEY_ID}
-    formSpacing={5}
-    ref={ref}
-  />
+  <div ref={ref}>
+    <SurveyComponent
+      {...props}
+      surveyId={PUBLIC_SURVEY_ID}
+      formSpacing={5}
+      ref={ref}
+    />
+  </div>
 ));
--- a/frontend/front/src/pages/Public/Pages/SurveyPage.js
+++ b/frontend/front/src/pages/Public/Pages/SurveyPage.js
@@ -57,18 +57,34 @@ export const SurveyPage = () => {
   );
 
   const handleAddSurveyVisit = useCallback(
-    async (responses, surveyId) => {
-      const recaptcha = await getReCaptchaToken();
-      return await addSurveyVisit({
-        responses,
-        recaptcha,
-        surveyId,
-        homeId: createHomeData?.id,
-        // TODO: probably remove this and handle on the back end
-        date: new Date().toISOString(),
+    async (responses, surveyId, activeHome) => {
+      const recaptcha = await getReCaptchaToken("create_survey");
+
+      const surveyAnswers = {};
+      Object.entries(responses).forEach(([key, value]) => {
+        surveyAnswers[key] = {
+          survey_question_id: key,
+          answer: value,
+        };
       });
+
+      const surveyVisit = await addSurveyVisit({
+        surveyVisit: {
+          survey_visit: {
+            home_id: activeHome.id,
+            surveyor_id: null,
+            survey_response_attributes: {
+              survey_id: surveyId,
+              completed: "true",
+              survey_answers_attributes: surveyAnswers,
+            },
+          },
+        },
+        recaptcha,
+      });
+      return surveyVisit;
     },
-    [addSurveyVisit, createHomeData?.id, getReCaptchaToken]
+    [addSurveyVisit, getReCaptchaToken]
   );
 
   const step = useMemo(() => {
--- a/frontend/front/src/pages/Surveyor/houseProfile/HouseProfile.js
+++ b/frontend/front/src/pages/Surveyor/houseProfile/HouseProfile.js
@@ -1,9 +1,4 @@
-import {
-  Alert,
-  Container,
-
-  Snackbar,
-} from "@mui/material";
+import { Alert, Container, Snackbar } from "@mui/material";
 import React, { useCallback, useMemo } from "react";
 import { useParams } from "react-router-dom";
 import { HeatPumpSlide } from "../../../components/HeatPumpSlide";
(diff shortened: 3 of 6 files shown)


frontend/front/src/pages/Public/Components/PublicSurvey.js
```diff
@@ -4,10 +4,12 @@ import SurveyComponent from "../../../components/SurveyComponent/SurveyComponent
 const PUBLIC_SURVEY_ID = "1";
 
 export const PublicSurvey = forwardRef((props, ref) => (
-  <SurveyComponent
-    {...props}
-    surveyId={PUBLIC_SURVEY_ID}
-    formSpacing={5}
-    ref={ref}
-  />
+  <div ref={ref}>
```
mattdelsordo, line 7: If you fixed the ref thing at the SurveyComponent level I don't think you need this div any more. Not sure what the behavior is if the ref is set to two different divs.

frontend/front/src/pages/Surveyor/houseProfile/HouseProfile.js
```diff
@@ -1,33 +1,42 @@
...
 const HouseProfile = () => {
   const { id: homeId } = useParams();
   const {
     data: homeData,
-    error: homesError,
-    isLoading: isHomesLoading,
+    error: homeError,
+    isLoading: isHomeLoading,
   } = useGetHomeQuery(homeId);
+  // const assignmentId = homeData?.assignment_id;
```
mattdelsordo, line 31: Remove these comments

frontend/front/src/pages/Surveyor/houseProfile/HouseProfile.js
```diff
@@ -41,27 +50,50 @@ const HouseProfile = () => {
...
   const submitSurvey = useCallback(
     async (responses, surveyId) => {
-      return await addSurveyVisit({
-        responses,
-        homeId,
-        surveyId,
-        // TODO: probably remove this and handle on the back end
-        date: new Date().toISOString(),
+      const surveyAnswers = {};
+      Object.entries(responses).forEach(([key, value]) => {
```
mattdelsordo, line 54: Since you have this in two places it might be good to factor it out into a shared util function

mattdelsordo (changes requested):
Looks good pending things that need to be ironed out

backend/app/views/survey_responses/_survey_response.json.jbuilder
```diff
@@ -1,4 +1,4 @@
 # frozen_string_literal: true
 
-json.extract! survey_response, :id, :survey_id, :created_at, :updated_at
+json.extract! survey_response, :id, :survey_id, :survey_visit_id, :created_at, :updated_at
```
LWest001, line 3: needed for finding which survey visit a survey response belongs to

frontend/front/src/pages/Public/Pages/SurveyPage.js
```diff
@@ -50,25 +51,23 @@ export const SurveyPage = () => {
...
 
   const handleAddSurveyVisit = useCallback(
-    async (responses, surveyId) => {
-      const recaptcha = await getReCaptchaToken();
-      return await addSurveyVisit({
-        responses,
+    async (answers, surveyId, homeId) => {
+      const recaptcha = await getReCaptchaToken("create_survey");
+      const surveyVisit = await addSurveyVisit({
+        surveyVisit: buildSurveyVisitData(answers, homeId, surveyId),
```
LWest001, line 65: Sends correctly formatted object to createSurveyVisit hook

frontend/front/src/pages/Public/Pages/SurveyPage.js
```diff
@@ -50,25 +51,23 @@ export const SurveyPage = () => {
 
   const handleCreateHome = useCallback(
     async ({ address }) => {
-      const recaptcha = await getReCaptchaToken();
-      createHome({ ...address, recaptcha });
+      const recaptcha = await getReCaptchaToken("create_survey");
+      const home = await createHome({ home: { ...address }, recaptcha });
+      return home;
```
LWest001, line 56: allows rtk-query to send recaptcha token header with createHome query

frontend/front/src/pages/Surveyor/houseProfile/HouseProfile.js
```diff
@@ -23,10 +24,12 @@ const STEP_THANKS = "PHASE_THANKS";
 
 const HouseProfile = () => {
   const { id: homeId } = useParams();
+  const { id: surveyorId } = useSelector(selectCurrentUser);
```
LWest001, line 27: Will function properly after merging #399

PR #400, opened 2023-06-19
User creation page

### LINKED ISSUE #356: Admin: create users
Admins should be able to create users. I think there's currently a screen for this but it needs the following:
- align inputs with actual `surveyor` data type
- `role` should be a dropdown or radio button to choose between `admin`, `surveyor`, `data_viewer`, and `data_editor`
- don't allow entering a password, generate it right before sending the request to the back end

On successful user creation, show the `email` + `password` combination.

### LINKED ISSUE #417: Whitelist DELETE for cors
forgot to do this

mattdelsordo:
Resolves #356
Resolves #417

### DIFF
--- a/backend/app/controllers/application_controller.rb
+++ b/backend/app/controllers/application_controller.rb
@@ -2,7 +2,7 @@
 
 class ApplicationController < ActionController::Base
   # before_action :authenticate_user!
-  
+
   # this fixed user creation but I have no idea WHY
   respond_to :html, :json
 end
--- a/backend/app/controllers/application_controller.rb
+++ b/backend/app/controllers/application_controller.rb
@@ -2,4 +2,7 @@
 
 class ApplicationController < ActionController::Base
   # before_action :authenticate_user!
+  
+  # this fixed user creation but I have no idea WHY
+  respond_to :html, :json
 end
--- a/frontend/front/src/pages/Admin/user/CreateNewUser.js
+++ b/frontend/front/src/pages/Admin/user/CreateNewUser.js
@@ -1,150 +1,209 @@
-import { Box, Button, TextField, Typography } from "@mui/material";
-import { Controller, useForm } from "react-hook-form";
+import {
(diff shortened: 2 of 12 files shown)


frontend/front/src/pages/Admin/user/CreateNewUser.js
```diff
@@ -1,149 +1,209 @@
...
+  }, [navigate]);
+
+  const onSubmit = useCallback(
+    async (data) => {
+      setIsUserLoading(true);
+
+      const password = generatePassword();
+
+      try {
+        const user = await createUser({
```
mattdelsordo, line 66: The lifecycle here is:
1. create devise user
2. user id to create surveyor
3. show the current user a modal with the email and auto-generated password to be passed along to the surveyor

Not 100% on how we should handle this, another option would be to give each user the same password and then let people reset the passwords individually (something I think we should do anyway).

mattdelsordo: Decision was to have Admin write passwords down and then have users change their passwords until we get email set up on the backend.

thiagobardini (approved):

backend/app/controllers/application_controller.rb
```diff
@@ -2,4 +2,7 @@
 
 class ApplicationController < ActionController::Base
   # before_action :authenticate_user!
+
+  # this fixed user creation but I have no idea WHY
+  respond_to :html, :json
```
mzagaja, line 7: @mattdelsordo What was the specific issue/test that this fixed? Seems like a client request was likely requesting something without sending the relevant JSON headers, or alternatively you needed one of the default Devise pages that does not have a JSON equivalent.

mattdelsordo: I think it was probably the latter. All I remember is that I was having tons of issues with devise endpoints responding with html forms instead of json responses.

PR #399, opened 2023-06-16
Various auth tweaks and fixes

mattdelsordo:
This should unblock the front end team! Although maybe someone else should test it out before this gets merged :sweat_smile:

PR #402, opened 2023-06-19
Surveyor view: Load My Account data

LWest001:
Loads account data for surveyor My Account page and displays it, with error handling and page loading.

To complete:
- [ ] Editing email address should [edit the user object](https://cfb-public.slack.com/archives/C03QW1GTJHW/p1687718831423539) for login purposes.


frontend/front/src/features/login/loginSlice.js
```diff
@@ -31,8 +34,8 @@ const loginSlice = createSlice({
     );
     // update state whenever user logs out successfully
     builder.addMatcher(apiSlice.endpoints.logoutUser.matchPending, (state) => {
-      state.user = null;
-      state.token = null;
+      state.user = initialState.user;
+      state.token = initialState.token;
```
mattdelsordo, line 38: You can probably just do `state = initialState`

frontend/front/src/pages/Admin/nav/Nav.js
```diff
@@ -8,12 +8,17 @@ import {
...
-import React from "react";
+import { Link, useNavigate } from "react-router-dom";
+import React, { useCallback } from "react";
 import { useLogoutUserMutation } from "../../../api/apiSlice";
 
 const Nav = () => {
+  const navigate = useNavigate();
   const [logout] = useLogoutUserMutation();
+  const handleLogout = useCallback(async () => {
+    await logout();
```
mattdelsordo, line 19: Are the mutation functions that rtk query returns actual Promises? I had to do `mutation().unwrap()` elsewhere to get it to actually await.

frontend/front/src/pages/Surveyor/account/Account.js
```diff
@@ -3,42 +3,69 @@ import React from "react";
...
+import { selectCurrentUser } from "../../../features/login/loginSlice";
+import { useGetSurveyorQuery } from "../../../api/apiSlice";
+import Loader from "../../../components/Loader";
+import CustomSnackbar from "../../../components/CustomSnackbar";
 
 const Account = () => {
-  const { firstName, lastName, email, address, phoneNumber } = useSelector(
-    (state) => state.account
-  );
+  const { id: userId } = useSelector(selectCurrentUser);
```
mattdelsordo, line 12: This won't work in practice, because the current user contains the _user_ id, but what you want is the _surveyor_ id, and they aren't guaranteed to be the same. #399 adds the surveyor email to the user state, and you should be able to use that to do like `/surveyors?email=${email}` to get the user data.

frontend/front/src/pages/Surveyor/account/Account.js
```diff
@@ -3,42 +3,69 @@ import React from "react";
...
-            </Button>
-          </Box>
-        </Grid>
-      </Grid>
+      {isAccountDataLoading ? (
+        <Loader />
+      ) : isAccountDataError ? (
+        <CustomSnackbar
+          open={isAccountDataError}
+          message="Error fetching Surveyors data."
```
mattdelsordo, line 33: "Surveyors" or "Surveyor's"? (or "surveyor"?)

frontend/front/src/pages/Surveyor/account/edit/EditAccount.js
```diff
@@ -38,168 +61,195 @@ const EditAccount = () => {
...
+          surveyor: {
+            firstname: values.firstName,
+            lastname: values.lastName,
+            phone: values.phoneNumber,
+            email,
+            street_address: values.address,
+          },
+        },
+      });
+      return updatedAccount;
```
mattdelsordo, line 78: You should make sure this updates the `user` object as well, otherwise if they change their email here they won't be able to use that to log in.

frontend/front/src/pages/Surveyor/account/edit/EditAccount.js
```diff
@@ -38,168 +61,195 @@ const EditAccount = () => {
...
+    },
+    [updateAccount, id]
+  );
+
   async function EditAccountForms(values) {
```
mattdelsordo, line 83: Just a style thing, this should be camel case not pascal case.

frontend/front/src/pages/Surveyor/account/edit/EditAccount.js
```diff
@@ -38,168 +61,195 @@ const EditAccount = () => {
...
+              fullWidth
+              label="Email"
+              variant="standard"
+              defaultValue={email}
+              {...register("email", {
+                required: {
+                  value: true,
+                  message: "Please Enter Email",
+                },
+                pattern: {
```
mattdelsordo, line 164: For all of these but especially the email, you might be able to be more concise by reusing the components for the SurveyComponent, HeatPumpTextField etc.

frontend/front/src/redux/store.js
```diff
@@ -9,7 +8,6 @@ import surveyorReducer from "../features/surveyor/surveyorSlice";
 export const createStore = (options) =>
   configureStore({
     reducer: {
-      account: accountReducer,
```
mattdelsordo, line 12: I approve of this removal

mattdelsordo (changes requested):

mattdelsordo:
Another thing that's probably out of scope for this particular thing is that we should allow surveyors to change their passwords on this screen.

frontend/front/src/pages/Surveyor/account/Account.js
mattdelsordo: Actually, should I just have it put the whole surveyor in the JWT?

frontend/front/src/pages/Admin/nav/Nav.js
LWest001: The mutation trigger (in this case, `logout`) returns a promise.
https://redux-toolkit.js.org/rtk-query/usage/mutations#frequently-used-mutation-hook-return-values

frontend/front/src/pages/Surveyor/account/Account.js
LWest001: What would be the advantage of storing the whole surveyor in the JWT? The solution in your previous comment makes sense to me.

frontend/front/src/pages/Surveyor/account/Account.js
LWest001: Probably "account data" is better in the context of this page. Copy and pasted this from another component and forgot to fix that detail!

frontend/front/src/pages/Surveyor/account/edit/EditAccount.js
LWest001: With HeatPumpTextField, I don't see that I can give it a `type="email"` attribute, or some of the other props like the `pattern` for regex testing. Maybe give it a `textFieldProps` prop or something.

frontend/front/src/pages/Surveyor/account/Account.js
mattdelsordo: I was thinking then you'd have the surveyor id itself

LWest001: note to self: select surveyor_id (or similar) instead of just id

frontend/front/src/pages/Surveyor/account/edit/EditAccount.js
LWest001: Meeting note: Maybe surveyor shouldn't have email at all, just user.

frontend/front/src/pages/Surveyor/account/Account.js
LWest001: @mattdelsordo When I curl `localhost:3000/surveyors?email=admin@heat.pump` or actually any params in that format (`/surveyors?id=1` etc.) I just get back the whole `surveyors.json` every time. We were running into a similar thing with getting unassigned homes with `/homes?assignment=null`. Maybe these param requests need to be explicitly set up?

frontend/front/src/pages/Surveyor/account/edit/EditAccount.js
LWest001: @mattdelsordo I'm not entirely sure how to update a user.

frontend/front/src/pages/Surveyor/account/Account.js
LWest001: I dug a bit - it's in the controller.
```
def search params
  params.permit(:param1, :param2, etc.)
```

LWest001: This is resolved in commit f34e333

mattdelsordo: Sorry to re-open this but with #399 you should be able to get the `surveyor_id` out of the user object directly! So you don't actually need to look up with the email after all.

frontend/front/src/pages/Surveyor/account/edit/EditAccount.js
mattdelsordo: I looked into this a bit, and it seems like this might be a bit more complicated than just doing a PATCH. I say for now, just don't let the user edit their email or password.

PR #405, opened 2023-06-21
Remove completed attribute from survey_response and calculate instead

mikelynch:
Instead of manually tracking whether a survey_response is completed, we determine that a home is completed when it has a survey_visit with a survey_response. This PR also adds a visited? method which sees if a home has a survey_visit. We do eager joins in the homes and assignments controllers to efficiently determine these properties on the show and index methods.


mikelynch:
One remaining issue @mzagaja : we couldn't figure out the syntax for the JBuilder for the homes collection on the assignment, so they still have attributes with `"viewed?": true, "completed?: false`, rather than having `viewed` and `completed` – this did work for the homes views though.

mikelynch:
We were trying something like this
```ruby
json.homes do
  json.array! assignment.homes.sort_by |h| { h[:visit_order] } do |home|
      json.extract! home, :id, :street_number, :street_name, :unit_number, :city, :state, :zip_code,
      :building_type, :visit_order
      json.visited home.visited?
      json.completed home.completed?
  end
end
```
but that definitely doesn't work!

mattdelsordo:
Any status updates on this change?

mikelynch:
We may want to revisit this as we've discussed some other changes (probably we will not track the “meta-questions” separately, but instead make them the first few questions of the survey). This would mean the `visited` property is not useful separately from the `completed` one, but for now the front-end team can start using this.

PR #406, opened 2023-06-21
Completed vs Incompleted on surveyor dashboard

### LINKED ISSUE #376: Surveyor view: Assignment accordion/Options button - Select Incompleted option should check boxes for incompleted homes

LWest001:
Closes #376.

**Note**: I originally set this up to calculate completed homes client-side. There will soon be a `completed` attribute exposed on the backend. See note at the end.

* ~~apiSlice: Add HomeCompleted tag, add getIsHomeCompleted query, createSurveyVisit invalidates HomeCompleted.~~
* AssignmentHome: New component abstracted from AssignmentUnit. ~~Uses a MUI Skeleton for loading Completed data.~~
* AssignmentUnit: Lots of cleanup (camelCasing, removing unnecessary abstractions). Set up -select incompleted- to work with ~~the getIsHomeCompleted query~~ the `home.completed` attribute. AssignmentHome abstracted out.
* ListView: Fix rendering logic and use standard error and loading components.

~~What do we think? Is this solution fine? It's not super fast but as far as I can tell it's the best we can do client-side.~~

Like PR #411, this change will work properly once we merge PR #405.


LWest001:
We can accomplish this more easily after merging #405 which will calculate if a home is completed and give it that attribute.

backend/app/controllers/homes_controller.rb
```diff
@@ -66,7 +66,7 @@ def home_params
   end
 
   def search_params
-    params.permit(:street_number, :street_name, :unit_number, :city, :state, :zip_code, :building_type, :assignment_id,
+    params.permit(:id, :street_number, :street_name, :unit_number, :city, :state, :zip_code, :building_type, :assignment_id,
```
mattdelsordo, line 69: This might not be necessary but otherwise looks good :+1:

mattdelsordo (approved):

PR #407, opened 2023-06-23
Basic production Docker build and run directions in README

### LINKED ISSUE #354: Create Dockerfile to build/host static website files
We can do the build + host with a multi-stage Dockerfile. End result can just be static files served by nginx

Maybe useful: https://docs.docker.com/build/building/multi-stage/

Rando-cal:
Resolves #354

### DIFF
--- a/frontend/front/README.md
+++ b/frontend/front/README.md
@@ -1,4 +1,4 @@
-lStart Guide By Dat Nguyen email: datguy7@gmail.com
+Start Guide By Dat Nguyen email: datguy7@gmail.com
 
 To run this Front-End only react application, you need VSCode IDE.
 download here: https://code.visualstudio.com/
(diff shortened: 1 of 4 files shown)


frontend/front/README.md
```diff
@@ -127,3 +127,44 @@ This section has moved here: [https://facebook.github.io/create-react-app/docs/d
...
+
+Goto /frontent/front and run yarn install
+
+Run 'npm run build' to make a production version
+
+Goto new /build folder
+
+Install Docker code if not already
+
+Make a Dockerfile and put the following in it:
```
mattdelsordo, line 148: Could you put this in a Dockerfile and commit that? Then we could reference it here instead of having folks make their own.

mattdelsordo (changes requested):

frontend/front/README.md
Rando-cal: @mattdelsordo I put the file in there plus a 79mb container. We need decide if we want to keep that in main....

mattdelsordo: The container should not be checked in to git, we should leave it up to the dev or the CI to build a container from the Dockerfile when needed.

Rando-cal:
I need to update and remove the TAR file. Will want to just have the README for directions.

frontend/front/README.md
Rando-cal: @mattdelsordo 

New commit.

I removed the TAR file and updated a script. It should install in Linux. Please read the README.md in docker-builds to see if it works for you. Thanks

frontend/front/docker-builds/Dockerfile
```diff
@@ -0,0 +1,11 @@
+# grab nginx image running on alpine linux
+FROM nginx:stable-alpine
+
+# update and add dependencies
+RUN apk update && apk upgrade && apk add openssl curl ca-certificates
+
+# copy contents of build folder to default folder in container
+COPY . /usr/share/nginx/html
```
mattdelsordo, line 8: Why not just do `COPY ../build /usr/share/nginx/html` ? Having to copy this into the build folder seems annoying.

mattdelsordo (approved):

Issue #421, opened 2023-07-12
Login Remember me box: Get it working


Rando-cal:

mattdelsordo:
This is currently commented out. The way I think we want this to work is that it should only store the JWT in localstorage if the Remember button is checked (its currently doing this 100% of the time). See the extraReducers in loginSlice.js

EfrainAD:
If the user refreshes the page, is the desired outcome that they are logged out?

mattdelsordo:
On one hand I feel like no, but then we would need to store the JWT which defeats the purpose. The correct way to handle this would probably be to have the button control the expiry time of the token somehow, but short of that I think logging users out on refresh is fine.

EfrainAD:
That is backend that would control this, since they choose when the token expires. From searching around, you can set it up with a refresh token for the 'remember me' that is used to re-login the user with a new session token. Though I am thinking they can just set the API to except a rememer me arg, then choose the expired time on that. 

Since the requirment is meet, I'll send the pull request for it.

mattdelsordo:
Resolved by #436

PR #436, opened 2023-07-22
Remember works as requested; if the user clicks 'Remember Me,' the to…

EfrainAD:
…ken will go into local storage. If not, then it simply stays in Redux issue #421

1. Uncommented out the remember me part of the login.
2. Changed `If (token)` to `if (meta.arg.originalArgs.remember)`. So if remember is `true` it will store it in the local storage.

### DIFF
--- a/frontend/front/src/features/login/loginSlice.js
+++ b/frontend/front/src/features/login/loginSlice.js
@@ -25,7 +25,7 @@ const loginSlice = createSlice({
           meta.baseQueryMeta.response.headers.get(AUTHORIZATION_HEADER);
         state.token = token;
         state.user = decodeJwt(token);
-        if (token) {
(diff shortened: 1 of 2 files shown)


frontend/front/src/features/login/loginSlice.js
```diff
@@ -25,7 +25,7 @@ const loginSlice = createSlice({
           meta.baseQueryMeta.response.headers.get(AUTHORIZATION_HEADER);
         state.token = token;
         state.user = decodeJwt(token);
-        if (token) {
+        if (meta.arg.originalArgs.remember) {
```
mattdelsordo, line 28: Slick solution :+1:

mattdelsordo (approved):

Issue #393, opened 2023-06-13
Remove completed attribute on survey_response and return with home


mikelynch:
Currently there is a `completed` attribute on `survey_response`. Given the anticipated survey length, surveyor will not return to ask more questions – if there are answers missing, they will stay that way. We should remove this attribute.

Instead, to determine if a home should be considered complete, we should look to see whether a home has a survey_response (via survey_visit) – if it does, we consider it completed.

PR #440, opened 2023-07-26
393: query survey_visits to mark things completed

mattdelsordo:
This isn't 100% what we want, since it can't check the survey_responses on the survey_visits. Should be removed pending #405

Related to #393


mattdelsordo:
This needs to be updated to use the real value of completed now

mattdelsordo:
PR now contains the proper changes for consuming `completed`!

PR #444, opened 2023-08-02
bug-fix: Fixed the Select dropdown components from cutting off survey…

EfrainAD:
Fixed the Select dropdown components from cutting off survey questions

   - Wrapped label text in a <div> element within <InputLabel> and obtained scroll height using the assigned reference. (useRef)
   - Dynamically adjusted <Select> height based on the calculated scroll height of the wrapping <div>.
   - Utilized the useEffect to refresh the comonent to get the accurate height from the <div>, since it would set height before Ref was ready.
   - Added an event listener to detect font size changes, to make the height change dynamic.
   - Applied padding on right to the label/div to prevent overlapping with the arrow icon.

### DIFF
--- a/frontend/front/src/components/SurveyComponent/HeatPumpDropdown.js
+++ b/frontend/front/src/components/SurveyComponent/HeatPumpDropdown.js
@@ -6,7 +6,13 @@ import {
   MenuItem,
   InputLabel,
 } from "@mui/material";
-import React, { useCallback, useEffect, useMemo, useRef, useState } from "react";
+import React, {
+  useCallback,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from "react";
 
 /**
  * A radio group to be used with react-hook-form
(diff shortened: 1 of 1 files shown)


frontend/front/src/components/SurveyComponent/HeatPumpDropdown.js
```diff
@@ -21,6 +21,21 @@ export const HeatPumpDropdown = ({
...
+  useEffect(() => {
+    const refresh = () => setResizeTrigger(prev => prev+1)
+    
+    refresh()
+
+    window.addEventListener('resize', refresh)
+      
+    return () => window.removeEventListener('resize', refresh)
+
+  })
```
mattdelsordo, line 36: Check the linting test, this needs a dependency array even if it's just `[]`.

Using an event listener and a useState to trigger a refresh here seems like a red flag to me. Would it be possible to have the useEffect depend on `labelFormRef.current?.scrollHeight` at least? I've looked into this 0% and this may be the only way but it seems like there ought to be a purely CSS way to handle this sort of thing.

mattdelsordo (changes requested):

mattdelsordo (approved):

Issue #415, opened 2023-06-28
Admin: link to survey_visit in home table if complete


mattdelsordo:
In the admin home table, we want to link to any complete survey visits alongside the homes, so you can see what responses we've gotten from a given home. Relies on #405

LWest001:
Assign to Efrain Davila when added to the organization

PR #447, opened 2023-08-08
Feature: The 'Completed' column is now presented as a button if the a…

EfrainAD:
…nswer is yes. When clicked, it navigates the user to the survey.

issue #415

   1. I utilized the useGetSurveyVisitsQuery to retrieve all survey data.

   2. Within the "Completed" column, I replaced the 'Yes' response with a button that facilitates user navigation to the survey.

   3. Next, I regenerated the house data by appending completed: true and survey_id: visit.id. Please note that currently, the built-in method for searching survey visits by quiries is on the to-do list, according to https://docs.google.com/document/d/1QAek7K-pvnCJe4wxm38QdPCbszEE-esw3w-tuhEtMDQ/edit. So, I couldn't directly retrieve it.

   4. Should any errors arise from survey visits, a CustomSnackbar is triggered to inform the user. However, the table still loads without interruption.


EfrainAD:
I thought a change that was pulled into main did the completed field already being generated, but I didn't seem to be. So maybe it not pulled in yet. So not sure how this request would change in response.

EfrainAD:
Build image is not passing and I have no idea how to get that to pass, or if I am supposed to. But I did run prettier and eslent on it.

frontend/front/src/pages/Admin/home/HomeTable.js
```diff
@@ -99,32 +122,61 @@ const HomeTable = () => {
...
+  } = useGetSurveyVisitsQuery();
+
+  let homesData = [];
 
-  if (isHomesDataLoading) {
+  if (surveyVisitsData && fetchedHomesData) {
+    homesData = Object.values(fetchedHomesData).map((home) => {
+      const visit = surveyVisitsData.find((visit) => visit.home_id === home.id);
+      if (visit) {
+        return { ...home, completed: true, survey_id: visit.id };
```
mattdelsordo, line 141: `home`s should have `completed` on them by default on `main` now, so you don't need to set it here. Also, for the record, I think that API document is out of date because the endpoint `/survey_visits?home_id={id}` totally works. It's still probably more appropriate here to get the whole survey_visit list since they have to be matched to the entire list of homes. That, or you could try adding `:survey_visit_ids` to the `json.extract!` line in `_home.json.jbuilder` in the back end code to have it return that data..

frontend/front/package.json
```diff
@@ -59,6 +59,7 @@
     ]
   },
   "devDependencies": {
+    "@babel/plugin-proposal-private-property-in-object": "^7.21.11",
```
mattdelsordo, line 62: What is this being used for?

frontend/front/src/pages/Admin/home/HomeTable.js
```diff
@@ -99,32 +122,61 @@ const HomeTable = () => {
...
+  const {
+    data: surveyVisitsData,
+    isError: isSurveyVisitsError,
+    isLoading: isSurveyVisitsDataLoading,
+  } = useGetSurveyVisitsQuery();
+
+  let homesData = [];
 
-  if (isHomesDataLoading) {
+  if (surveyVisitsData && fetchedHomesData) {
```
mattdelsordo, line 137: 100% you want to wrap this in a `useMemo`, because otherwise it's going to recompute all of this on each rerender, regardless of whether the data has changed.

mattdelsordo (changes requested):

mattdelsordo:
Oh, didn't notice your above comments. I'm willing to bet both the Build Image failure and the lack of the completed field are because your fork is out of sync with the original version of `main`. Not 100% sure how to resolve that.

frontend/front/package.json
EfrainAD: The eslent? message said to add that to pass the check. Said that it was most likely not to be fixed.

EfrainAD:
Ok This should be up and run

EfrainAD:
Fixing my mistake, reopening.

EfrainAD:
Ok, now been updated with how the backend works now. I did the merge conflits. I did the yarn file conflits, I admit I'm hoping it is correctly done.

frontend/front/src/pages/Admin/home/HomeTable.js
EfrainAD: With the code changes, this is back to 
if (isHomesDataLoading) {
    return <Loader />;
}
I am thinking this is fine as is?

frontend/front/src/pages/Admin/assignment/AssignTable.js
```diff
@@ -175,6 +175,58 @@ const AssignTable = () => {
         }`;
       },
     },
+    {
```
mattdelsordo, line 178: It looks like here you're adding new surveyor/completed columns instead of editing the existing ones.

frontend/front/src/pages/Admin/home/HomeTable.js
```diff
@@ -33,10 +36,19 @@ const HomeTable = () => {
 
   const handleHomeLink = (home) => goToBreadcrumb("home", home);
 
+  const handleUserLink = (home) => {
+    navigate("/admin/survey/visit/" + home.survey_visit_ids[0]);
```
mattdelsordo, line 40: Use string interpolation here, e.g.
```
`/admin/survey/visit/${home.survey_visit_ids[0]}`
```

frontend/front/src/pages/Admin/home/HomeTable.js
```diff
@@ -76,7 +88,20 @@ const HomeTable = () => {
     {
       field: "completed",
       headerName: "Completed",
-      renderCell: (params) => (params.row.completed === true ? "Yes ✅" : "No"),
+      renderCell: (params) =>
+        params.row.completed === true ? (
```
mattdelsordo, line 92: This can probably just be `params.row.completed` actually - but that wasn't your addition so whatever

frontend/front/src/pages/Admin/home/HomeTable.js
mattdelsordo: Also, not sure how we should be handling survey visits > 1, but I think your approach here is correct.

mattdelsordo (changes requested):

frontend/front/src/pages/Admin/assignment/AssignTable.js
EfrainAD: I don't recall this, I think I took the Button code from here. I might have saved, and prettier formatted this? I tried to see what was "deleted', but it's not showing me that.

frontend/front/src/pages/Admin/home/HomeTable.js
EfrainAD: I have trouble hearing in the meetings and wish there was a source of truth on this. Are they still allowing more then one survey?

Maybe if there more then one have a collapsible section that will drown down? We may not be able to do that with the DataGrid mui component, but something on those lines.

frontend/front/src/pages/Admin/home/HomeTable.js
EfrainAD: I changed it.

frontend/front/src/pages/Admin/home/HomeTable.js
EfrainAD: Changed to string interpolation.

frontend/front/src/pages/Admin/home/HomeTable.js
mattdelsordo: Yeah you can just leave it as is

frontend/front/src/pages/Admin/assignment/AssignTable.js
mattdelsordo: In that case could you remove this?

EfrainAD: Ok, removed. 

I think I know how I mistakenly put this there. I have an upstream clone now. Though I made it for a different reason, it should help me be more careful.

frontend/front/src/pages/Admin/home/HomeTable.js
EfrainAD: This is no longer there. 

Thanks for letting me now, I checked in postman. I'll just do that from now on.

PR #448, opened 2023-08-08
Frontend issue413 admin home profile

mcarre20:
fix home profile, will still need to add a route to "view survey button"


mattdelsordo (approved):
Works great! Go ahead and merge.

Some notes that are maybe out of scope of this particular thing:
- we might want a success message for when a save is successful
- the View Survey button doesn't do anything, but we have everything in place to get that hooked up. If the home has `completed === true`, display the button and link to the survey_visit that's associated with the home, otherwise hide the button. We might also want to reconsider the UI here if there's >1 survey_visit

PR #458, opened 2023-08-21
Admin viewing completed survey answers

EfrainAD:
I found that the Survey Structure only has one entry regardless of how many surveys. So I made changes on files I worked on to pass nothing and changed the apiSlice to have default as 1 when it's used to query.


EfrainAD:
Test UI fail is fixed in one of my other pull request. 

babel-preset-react-app is part of the create-react-app project, which
is not maintianed anymore. It is thus unlikely that this bug will
ever be fixed. Add "@babel/plugin-proposal-private-property-in-object" to
your devDependencies to work around this error. This will make this message
go away.

frontend/front/src/components/SurveyComponent/SurveyComponent.js
```diff
@@ -250,17 +252,17 @@ const SurveyComponent = ({
...
 // makes sure data fetching happens BEFORE the form is loaded, so that the form hook can be initialized with the correct default data
 const SurveyComponentWrapper = forwardRef((props, ref) => {
-  const { defaultData, surveyId, style, activeHome } = props;
+  const { defaultData, style, activeHome } = props;
 
   const {
     data: surveyStructure,
     isError: isSurveyError,
     isLoading: isSurveyLoading,
-  } = useGetSurveyStructureQuery(surveyId);
```
mattdelsordo, line 259: This is the wrong approach, in the future we plan to have more than one surveyId and it needs to be configurable here (although it will probably stay hard-coded somewhere).

frontend/front/src/api/apiSlice.js
```diff
@@ -137,7 +137,7 @@ export const apiSlice = createApi({
       ],
     }),
     getSurveyStructure: builder.query({
-      query: (id) => ({
+      query: (id = 1) => ({
```
mattdelsordo, line 140: Again, I don't think this is a great way to handle this.

frontend/front/src/pages/Admin/home/SurveyVisitProfile.js
```diff
@@ -71,9 +73,9 @@ const SurveyProfile = () => {
       </Typography>
       {surveyVisit && houseData ? (
         <AdminSurvey
-          defaultData={surveyVisit.responses}
+          defaultData={surveyAnswers}
           activeHome={houseData}
-          surveyId={surveyVisit.surveyId}
+          surveyId={surveyVisit.id}
```
mattdelsordo, line 78: `surveyVisit.id` is the ID of the survey_visit, but it needs to be the ID of the _survey_ as in the list of questions, which I think would be `surveyVisit.survey_response.survey_id` here.

frontend/front/src/components/SurveyComponent/SurveyComponent.js
```diff
@@ -250,17 +252,17 @@ const SurveyComponent = ({
...
+  } = useGetSurveyStructureQuery();
 
   const formDefault = useMemo(() => {
     if (defaultData) {
-      return defaultData;
+      return buildDataFromSurveyAnswers(defaultData);
```
mattdelsordo, line 265: I think you should call this one level up, when you pass the data into this from `SurveyVisitProfile`. `defaultData` is already passed in in the correct format on other pages (at least it should be, we need more tests...) so we should assume that it's always in the format we're looking for by the time it gets here.

EfrainAD: Done. Pushed up. I put in a useMemo. I have never used this before and I still need to do more reading on it. Just let me know if wrong.

frontend/front/src/pages/Admin/home/SurveyVisitProfile.js
EfrainAD: I need learn to keek all this servey names straight. Ok I didn't see anything at the time that was close to matching Id and thought that what was ment. Though it was in the name 'surveyId'. I pushed up that change.

frontend/front/src/components/SurveyComponent/SurveyComponent.js
EfrainAD: Ok, I thought maybe it was changed to only have one question survey or something. How do I go about this for now then? This will break it if I pass it anything other then a 1. First survey will be the only one that shows. Or are we ok with that?

frontend/front/src/pages/Admin/home/SurveyVisitProfile.js
```diff
@@ -14,18 +14,29 @@ import { SurveyError } from "../survey/SurveyError";
...
 
   const { data: houseData, error: houseError } = useGetHomeQuery(
-    surveyVisit?.homeId,
+    surveyVisit?.home_id,
     { skip: !surveyVisit }
   );
+
+  const surveyAnswers = useMemo(
+    () =>
+      surveyVisit?.survey_response
```
mattdelsordo, line 32: You can probably make this `surveyVisit?.survey_response?.survey_answers` like you have it on the next line, but you did do this correctly :+1:

frontend/front/src/components/SurveyComponent/SurveyComponent.js
EfrainAD: Got it. I didn't update what that meant in my head. I'll get to fixing it tonight.

EfrainAD: Done.

frontend/front/src/pages/Admin/home/SurveyVisitProfile.js
EfrainAD: Sorry, I don't understand. If you mean the way it is formatted, prettier is formatting it that way.

mattdelsordo: The formatting's fine, I mean you're checking that `surveyVisit?.survey_response` exists in the conditional, but you can make it like an ounce more efficient by changing the check to `surveyVisit?.survey_response?.survey_answers`, to avoid the call to buildDataFromSurveyAnswers if `survey_answers` is null or something. But it's fine as is!

mattdelsordo (approved):
Looks good! Go ahead and merge it!

PR #459, opened 2023-08-21
Public survey feature flag

mattdelsordo:
* Hides the public survey unless `REACT_APP_PUBLIC_SURVEY_ENABLED=true`
* Added example .env file for the front end
* Fixed the failing snapshot test


.github/workflows/test.yml
```diff
@@ -67,7 +67,7 @@ jobs:
       - name: Git checkout
         uses: actions/checkout@v3
       - name: install dependencies
-        run: yarn
+        run: yarn --immutable
```
mattdelsordo, line 70: The snapshot was failing because here it was not just installing but updating dependencies, and a new version of MUI had a different ID for some component.

PR #460, opened 2023-08-22
Create home

mcarre20:
create new home from admin page.


frontend/front/src/pages/Admin/home/CreateNewHome.js
```diff
@@ -1,28 +1,47 @@
...
 
 import { withAdminPrefix, ADMIN_HOME } from "../../../routing/routes";
 
 const CreateNewHome = () => {
   const navigate = useNavigate();
-  const { handleSubmit, control } = useForm({
+  const { handleSubmit, control, reset } = useForm({
     defaultValues: {
       streetNumber: "",
-      address: "",
```
mattdelsordo, line 14: The back end uses different casing for these fields (`street_number` instead of `streetNumber` etc.) so I'm not sure that this will work.

frontend/front/src/pages/Admin/home/CreateNewHome.js
```diff
@@ -88,14 +106,31 @@ const CreateNewHome = () => {
...
+              />
+            )}
+          />
+          {isError && (
+            <Alert severity="error" sx={{ my: 1 }}>
+              Error Saving home
+            </Alert>
+          )}
           <Box pt={5} textAlign="right">
             <Button
```
mattdelsordo, line 128: It might be a good idea to disable this button if the mutation is loading.

mattdelsordo (changes requested):

Issue #429, opened 2023-07-19
Make sure map opens in phone app instead of browser tab


mattdelsordo:
* might need a long-press on the phone side (at least on iOS)
* might need a special URI format
* might need to be a real <a> link instead of a button calling `window.open`

LWest001:
@mattdelsordo, do you mind explaining your trick for running the backend when using mobile?

mattdelsordo:
Step 1: figure out what your local IP is (`192.168.x.x`, not `localhost`)
Step 2: run the dev server with `REACT_APP_API_URL={your IP}` (or just manually change the baseUrl in apiSlice.js)
Step 3: navigate to `{your IP}: 3001` on your phone when connected to the same wifi network as the computer running the dev server

mattdelsordo:
I've confirmed that this works at least in Android + Chrome. It does NOT work in Android + Firefox.

mattdelsordo:
@LWest001 "de"-confirmed that this does NOT work on iOS, so there's work to be done there.

mattdelsordo:
- We're using window.open rather than a proper link, which may be a part of the problem. 
- Specify to use a specific recommended browser for best experience
- In general: we should be using react-router Links instead of <a>s/etc.

mattdelsordo:
Potentially helpful link: https://developers.google.com/maps/documentation/urls/ios-urlscheme
At minimum, I think we want to have this working in iOS+Safari+Apple Maps and Android+Chrome+Google Maps (already working)

LWest001:
### Helpful links
[Detecting user's browser/OS with browser APIs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Browser_detection_using_the_user_agent)
[Google Maps URL Scheme for iOS](https://developers.google.com/maps/documentation/urls/ios-urlscheme)

PR #475, opened 2023-09-30
#429 - feature: If someone is using iOS (iPhone, iPod, iPad), it will open t…

EfrainAD:
- I tested this on iphone SE 2nd, iphone 8+
- I do not have an android to check with.
- I wan't able to find anything on getting apple to use the waypoints, so I made it up and it worked.

feature: If someone is using iOS (iPhone, iPod, iPad), it will open the apple maps app instead of google maps.

1. Fixed a bug that one of the houses was getting cut off from the list of waypoints.
2. Use ‘navigator.userAgent’ to know if the user is using iOS or not
3. The url and the waypoints variable are now formatted differently for the two cases. (iOS/!iOS)
4. On Apple Maps, the waypoints can get removed based the the mode you pick. Therefor I had removed it, but then added it back. We will need a person to test in person if walk mode removes the waypoints or not.

### DIFF
--- a/frontend/front/src/pages/Surveyor/map/mapUtils.js
+++ b/frontend/front/src/pages/Surveyor/map/mapUtils.js
@@ -7,7 +7,7 @@ export const generateMapsLink = (locations) => {
       .map((location) => `&daddr=${location.latitude},${location.longitude}`)
       .join("");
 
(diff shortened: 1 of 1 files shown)


mattdelsordo (approved):
Looks good to me, I've tested this in chrome+android before and it works fine there.

PR #480, opened 2023-10-04
Implement Language Preference Handling

thiagobardini:
### What does this PR do?
- Initializes language preference from localStorage, or sets it to 'en-us' by default.
- Adds a URL query parameter 'langPref' for public routes to store the language preference.
- Removes the 'langPref' query parameter for non-public routes, but retains the user's language preference in localStorage so the backend can use it if necessary.
- Stores the language preference in localStorage to maintain the user's choice across navigation.

### How to test?
- Navigate through the app and change the language preference to see if it is being stored and applied correctly.

### Queries:
https://www.bostonhpa.org/public?langPref=en-us
https://www.bostonhpa.org/public?langPref=fr-ht
https://www.bostonhpa.org/public?langPref=pt-br
https://www.bostonhpa.org/public?langPref=es-xm

#### Note: 
- I need to figure out how to exclude the URL when it doesn't start with /public. I can do this in the external API configuration, and I believe this code will handle it. I'll work on it this week.

### Screenshots
https://imgur.com/Y2pdsCy

## Live Demo
https://bhpa-ethimyit1-thiagobardini.vercel.app/public?langPref=en-us

### DIFF
--- a/frontend/front/src/pages/Public/PublicContainer.js
+++ b/frontend/front/src/pages/Public/PublicContainer.js
@@ -1,7 +1,7 @@
-import React from "react";
+import React, { useState, useEffect } from "react";
 import { ThemeProvider, Box, Stack } from "@mui/material";
 import { responsiveTheme } from "./Assets/theme";
-import { Route, Routes } from "react-router-dom";
+import { Route, Routes, useLocation } from "react-router-dom";
 import Home from "./Pages/Home/Home";
 import Navbar from "./Layout/Navbar";
 import Footer from "./Layout/Footer";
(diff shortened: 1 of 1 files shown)


thiagobardini:
Sorry, I haven't yet figured out how to use the external API. Please don't merge it yet, I'll let you guys know until at the end of week. @mattdelsordo @mzagaja

thiagobardini:
Hi, I'm closing this ticket because we've chosen to run a manual test on the Public static pages. We'll be using the react-i18next library for this implementation. 
Here is the ticket: https://github.com/codeforboston/urban-league-heat-pump-accelerator/issues/486

Issue #449, opened 2023-08-09
Filter out null/empty answers from submitted surveys


mattdelsordo:
Check form content before the submission:
- if an answer is empty, remove it
- make sure all fields default to null/empty and not the first option
- if NO questions have truthy values, there should be NO survey response

PR #482, opened 2023-10-09
#449: Filter out empty answers before submitting the survey

EfrainAD:
1. The buildSurveyVisitData function now includes an 'if (answers[key])' statement. So the survey data builds only for questions that have received a response. For example, {1: ‘’, 2: ‘yes’, 3: ‘’} becomes { 2: { survey_question_id: 2, answer: ‘yes'}}.

Note: This created an issue that when the survey only received responses for question 2, and the rest of the survey answers were undefined. And mui wornings went crazy in a loop.

2. The buildDataFromSurveyAnswers function then was modified to handle the missing data that came back. A numberOfQuestions parameter was added to it. For example, if the survey received only answer 2, 1, 3,.... was constructed as {1: ‘’, 2: ‘yes’, 3: ‘’.....}

3. I needed to get the number of survey quetions, so in SurveyVisitProfile.js, I integrated the useGetSurveyStructureQuery to dynamically retrieve the number of survey questions. I didn't want to hardcoding this value.

### DIFF
--- a/frontend/front/src/pages/Admin/home/SurveyVisitProfile.js
+++ b/frontend/front/src/pages/Admin/home/SurveyVisitProfile.js
@@ -3,6 +3,7 @@ import React, { useCallback, useMemo } from "react";
 import {
   useDeleteSurveyVisitMutation,
   useGetHomeQuery,
+  useGetSurveyStructureQuery,
   useGetSurveyVisitQuery,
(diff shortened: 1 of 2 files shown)


mattdelsordo (approved):
looks good to me :+1:

Issue #495, opened 2023-10-23
Homes lose street_number after canonicalization


LWest001:
Homes are stripped of their `street_number` attribute once canonicalized. While `status` is uncanonicalized, the street number is present, but it is an empty string when `status` becomes canonicalized.
Eg. submitting 11 Brookford St, Boston, MA 02125.

Server response before canonicalized:
```
{
    "id": 7441,
    "street_number": "11", 
    "street_name": "brookford st",
    "unit_number": "",
    "city": "boston",
    "state": "MA",
    "zip_code": "02125",
    "status": "uncanonicalized",
    ...rest of Home
}
```

After canonicalized:
```
{
    "id": 7441,
    "street_number": "",
    "street_name": "brookford st",
    "unit_number": "",
    "city": "boston",
    "state": "MA",
    "zip_code": "02125",
    "status": "canonicalized",
    ...rest of Home
}

mikelynch:
Somewhat related – we may need to get our own (updated) Pelias instance.

PR #497, opened 2023-10-25
Resolve Homes lose street_number after canonicalization #495

mzagaja:
* Update code to read Pelias data correctly.
* Add test for failing case.
* Fix matching algorithm.


mikelynch (approved):
LGTM

Issue #471, opened 2023-09-20
Admin + Surveyor page to view completed surveys


mattdelsordo:
In general, we don't have a component for viewing completed surveys. Probably makes sense to reuse the existing survey component, but make it read only.

Admin page:
- link in individual assignment page is broken
- should also be linked from home table but that is also missing

Surveyor page:
- we need some way to view complete surveys IN ADDITION to filling out new ones

EfrainAD:
The surveyor needs a way to view complete surveys. 
How do you want this done? Surveyor can see if it's completed or not, but not see the answer as it is. 
1. Since we want them to be able to create multiple survey per house, how do you want to handle it when they click on it. I guessing two options. 1. If they want to see an old survey (and pick one if their is more then one), or 2. complete a new survey as it set now.
3. Do we want surveyors to be able to edit it?

"IN ADDITION to filling out new ones"
I don't know what this means, I misread and added a way to filter out completed homes. Maybe that's what you meant?

mattdelsordo:
Re: how this should be done, we don't have any UX for this, so it's up to you how you want to design it. We definitely need some sort of a way to select from a list if there are multiple completed surveys.

As for the surveyor requirements, currently surveyors can fill out surveys for a home, but they can't view surveys that they've already submitted. I think we've decided in the past that surveyors should not be editing surveys and should just submit another one if it comes to it. Filtering out completed homes was not part of this but I see how that could be useful so you can probably hang on to that.

mattdelsordo:
I'm going to re-open this to add the `SurveyLink` component + read-only survey page to the surveyor view.

Rando-cal:
Andrewy-gh will take

PR #499, opened 2023-10-31
Efrain #471 adimin assignment page link to survey

EfrainAD:
I’m sorry this is late. I don’t have my computer, so I can’t test it with the backend change. I was waiting for the backend to run this. For personal reasons I had leave home, and don’t have my setup. I’m hoping someone will pull it down quickly to make sure it doesn’t break. 

This is for the admin table and at the survey table. (Or something like that. There are two locations that have the link.) I’m on an iPad, so I’m leaving out how I normally comment these. There are the commits.

### DIFF
--- a/frontend/front/src/components/SurveyLink.js
+++ b/frontend/front/src/components/SurveyLink.js
@@ -1,0 +1,44 @@
+import { Button, Menu, MenuItem } from "@mui/material";
+import { useState } from "react";
+import { useNavigate } from "react-router-dom";
+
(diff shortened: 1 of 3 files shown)


mattdelsordo (approved):

mattdelsordo:
Actually - could you add the `SurveyLink` component somewhere in the `AssignmentUnit` component as well, so that surveyors have access to this? If you don't have the time feel free to just merge this and I can add it in.

PR #506, opened 2023-11-21
Let surveyors view past surveys in read-only mode, display embedded map on survey address card

mattdelsordo:
image
image


frontend/front/src/components/SurveyComponent/SurveyComponent.js
```diff
@@ -208,7 +210,7 @@ const SurveyComponent = ({
                     key={`q${q.id}`}
                     control={control}
                     name={`${q.id}`}
-                    label={q.text}
+                    label={q.question}
```
mattdelsordo, line 213: The field for the survey question "question" changed from text to question at some point...

PR #548, opened 2024-04-06
Frontend minorbugfix

mcarre20:
-fix lint issues
-fix text-field id conflict in login page


mcarre20:
fix minor bugs on public page

PR #635, opened 2024-10-10
Update csv

Andrewy-gh:
- Update csv for survey questions

Issue #633, opened 2024-10-09
Change SurveyAnswer answer attribute to be a String with array: true


mzagaja:
Update SurveyAnswer to be a String with `array: true` set.

lfilmeyer:
https://github.com/codeforboston/urban-league-heat-pump-accelerator/commit/a3c95f6a8152b701f9e71c3d6964f61d7ebf89b1

PR #636, opened 2024-10-14
Add new checkbox component

Andrewy-gh:
- Setup checkbox component

Will implement into SurveyComponent.js after #633 and #635 is closed


mzagaja (approved):
Randy gave it a 👍🏻 so we know who to blame if it doesn't work.

Issue #641, opened 2024-10-18
Import real assignments in seed_importer


lfilmeyer:
Currently in `seed_importer` we import assignments from `test_assignments.csv`, which contains 7 fake assignments.  We then import real homes from `clustered_ordered_parcels.csv`, which includes the field `cluster` which becomes the `assignment_id` on a home.  These homes then reference a non-existent assignment (ex. the home for 104 WELLES AV DORCHESTER, 02124 has cluster number `212` which doesn't exist in `test_assignments.csv`).  We think that maybe `cluster_centroids_with_regions.csv` in the `data-analysis` folder could serve as the real assignments CSV to import from, but we should test this out.

PR #643, opened 2024-10-21
641 Import real assignments in seed_importer

lfilmeyer:
[See issue here](https://github.com/codeforboston/urban-league-heat-pump-accelerator/issues/641)

We're currently importing fake assignments. This PR changes `seed_importer` to import assignments from `cluster_centroids_with_regions.csv`.  When importing homes, it then looks up the correct assignment by `cluster_id` and associates it to the home.

Acceptance criteria:
* `seed_importer_spec` passes
* Dropping the database & re-seeding it locally results in correct number of assignments (145) & correct home <-> assignment associations.

Screenshot 2024-10-21 at 1 47 30 PM


backend/db/migrate/20241021114918_add_cluster_id_to_assignments.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class AddClusterIdToAssignments < ActiveRecord::Migration[7.1]
+  def change
+    change_table :assignments, bulk: true do |t|
+      t.integer :cluster_id
+
+      t.string :latitude
+      t.string :longitude
```
lfilmeyer, line 9: I added latitude & longitude to assignments instead of putting the data into `geocode` because I don't think we do anything with `geocode` at the moment, and having them separate seems cleaner & easier to use.  I would proposed dropping `geocode`.

I would also propose dropping `group` from `assignments` since I don't think we are using it or know how to populate it.

It's also worth mentioning that the `Assignments` admin view is a little lacking.  Without a map view, it's not necessarily clear to someone what the best way to assign an assignment is (& assignments close in id are not necessarily close to one another geographically)

backend/lib/seeds/test_homes.csv
```diff
@@ -1,103 +0,0 @@
-street_number,street_name,unit_number,city,state,zip_code,building_type,assignment_group,visit_order
```
lfilmeyer, line 1: This & `test_seeds` were removed in https://github.com/codeforboston/urban-league-heat-pump-accelerator/commit/f097fa46ca78a8bec6d85c6f449b6a0ab36f17f2

lfilmeyer:
As an aside, I wanted to sanity check the data in `cluster_centroids_with_regions.csv` which came from the [data-analysis folder ](https://github.com/codeforboston/urban-league-heat-pump-accelerator/blob/main/data-analysis/cluster_centroids_with_regions.csv) since it had been a while & the people who generated the data weren't around to vouch for it.  Things seemed good 👍. There are no homes with an invalid cluster id & there are no empty assignments. When mapping the data, homes (the small dots) were assigned a reasonable looking cluster location (big dots).
Screenshot 2024-10-21 at 2 01 08 PM

Screenshot 2024-10-21 at 2 01 18 PM

PR #650, opened 2024-11-27
632 add checkbox to survey

Andrewy-gh:
- Update csv - removed if clauses from question, add checkbox response type
- Add checkbox response type to backend
- Add checkbox component to survey component
- Add utility function to send array values to backend
- Update ids on radio and checkbox to fix warnings


Andrewy-gh:
Have an updated pr, closing this one

PR #670, opened 2025-03-24
Minor cleanup from ruby upgrade & nokogiri CVE update

lfilmeyer:


.devcontainer/devcontainer.json
```diff
@@ -19,15 +19,12 @@
...
 	},
-    "onCreateCommand": "backend/bin/setup",
+  "onCreateCommand": "backend/bin/setup",
 
 	// Use 'forwardPorts' to make a list of ports inside the container available locally.
 	// This can be used to network with other containers or the host.
 	// "forwardPorts": [3000, 5432],
 
-	// Use 'postCreateCommand' to run commands after the container is created.
-	// "postCreateCommand": "backend/bin/rake db:seed",
```
lfilmeyer, line 29: I'm actually not sure why this command (which was uncommented prior to ruby upgrade) worked before. In the `onCreateCommand`, we run `backend/bin/setup`, which runs `bin/rails db:prepare`, which includes seeding the database.  So with this command uncommented, we were attempting to seed the database twice, resulting in the error

```
ActiveRecord::RecordInvalid: Validation failed: Display order has already been taken
```

PR #678, opened 2025-04-08
653 add survey to home distance

poster515:


backend/app/models/csv_export_helper.rb
```diff
@@ -70,14 +73,19 @@ def self.home_assignment_hash(home)
   private_class_method :home_assignment_hash
 
   def self.survey_visit_metadata_hash(survey_visit)
+    distance_miles = Geocoder::Calculations.distance_between(
```
lfilmeyer, line 76: this will run the distance calculation on every survey visit every time we export.  maybe we could instead calculate it one time with the survey visit is created and store it on the survey visit

backend/app/models/csv_export_helper.rb
```diff
@@ -23,7 +25,8 @@ class CsvExportHelper
     survey_visit_longitude: 'Survey Visit Longitude',
     survey_visit_time: 'Survey Visit Time',
     surveyor_id: 'Surveyor ID',
-    surveyor_name: 'Surveyor Name'
+    surveyor_name: 'Surveyor Name',
+    survey_distance_miles: 'Survey Distance'
```
lfilmeyer, line 29: i might give this a more descriptive name like "survey_visit_to_home_distance" so that it's clear what it's the distance between

backend/spec/models/csv_exporter_spec.rb
```diff
@@ -1,27 +1,43 @@
...
+        [TestConstants::HOME_LAT, TestConstants::HOME_LONG],
+        [TestConstants::SURVEY_LAT, TestConstants::SURVEY_LONG]
+      )
+
       expected_headers = "#{expected_standard_headers},\"1. Do you want a heat pump?\"\n"
       expected_survey_visit = "#{@survey_visit.id},#{@home1.id},Yes,No,#{@assignment.id},#{@surveyor.id}," \
-        'Luna Peters,1,Broadway,106,Cambridge,MA,02139,42.32603453,-71.08999264,42.3281053,-71.08229235,' \
-        "2025-02-13 12:30:20 -0500,#{@surveyor.id},Luna Peters,Yes I would love a heat pump\n"
-      expected_home = ",#{@home2.id},Yes,,,,,1,Broadway,106,Cambridge,MA,02139,42.32603453,-71.08999264,,,,,,\n"
+        "Luna Peters,1,Broadway,106,Cambridge,MA,02139,#{TestConstants::HOME_LAT},#{TestConstants::HOME_LONG}," \
```
lfilmeyer, line 36: not crazy what you're doing here, but i tend to like to make the tests as dumb as possible to reduce possibly making bugs in them / as close to the CSV output & would undo the TestConstants

lfilmeyer (commented):
great first pass!!

added comments for things discussed during hack night

PR #688, opened 2025-05-03
Frontend add survey mode param

Andrewy-gh:
- add ```survey_mode``` param to backend query and update query calls
- update props for ```SurveyorSurvey```, ```PublicSurvey``` and ```AdminSurvey``` to use their respective language and survey mode settings
- add tests for langPref and surveyMode to ```SurveyComponent```
- remove second arg form ```buildDataFromSurveyAnswers``` util function and remove unneeded ```useGetSurveyStructureQuery``` query for ```SurveyVisitProfile``` as the length of query result is only used. ```useGetSurveyVisitQuery``` can provide same information.


mzagaja:
I think this is good to merge since Randy trusts you.

PR #691, opened 2025-05-05
Validate value of language code on survey response

lfilmeyer:
With this PR: https://github.com/codeforboston/urban-league-heat-pump-accelerator/commit/89f63aa84b5e3c27a4e839be1c590cdade097e38 where we added language code to survey response, we discussed that we should validate the value of language code.

### DIFF
--- a/backend/app/models/survey_response.rb
+++ b/backend/app/models/survey_response.rb
@@ -1,8 +1,12 @@
 # frozen_string_literal: true
 
 class SurveyResponse < ApplicationRecord
+  SUPPORTED_LANGUAGE_CODES = %w[en es ht pt].freeze
+
   belongs_to :survey
   belongs_to :survey_visit
   has_many :survey_answers, dependent: :destroy
   accepts_nested_attributes_for :survey_answers
+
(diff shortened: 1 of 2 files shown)


backend/app/models/survey_response.rb
```diff
@@ -1,8 +1,12 @@
...
 
 class SurveyResponse < ApplicationRecord
+  SUPPORTED_LANGUAGE_CODES = %w[en es ht pt].freeze
+
   belongs_to :survey
   belongs_to :survey_visit
   has_many :survey_answers, dependent: :destroy
   accepts_nested_attributes_for :survey_answers
+
+  validates :language_code, inclusion: { in: SUPPORTED_LANGUAGE_CODES }, allow_nil: true
```
lfilmeyer, line 11: `allow_nil: true` means that submitting surveys will still work until the front end updates things on their end to send the language code on the survey visit request.  not sure if we care strongly about removing this long term.

mzagaja: Probably not a huge deal as long as the FE crew doesn't forget to do their part.

mzagaja (approved):

PR #697, opened 2025-06-17
682 assignment table perf fix

Andrewy-gh:
- Added client-side pagination to reduce the initial number rows 
- Removed React state and use the DataGrid's [GridApi](https://v6.mui.com/x/api/data-grid/grid-api/). This seemed to speed up row selection.
- Add an isFetching flag to show loading spinner when making assignment updates/removals.
- Modify the add/remove handlers so that surveyors already assigned are not added again, and only those assigned to the selected assignments can be removed.
- Move `SurveyorSelect.js` into it's own file, now we are preventing `AssignTable.js` to be re-rendered on every surveyor change.

### DIFF
--- a/frontend/front/src/pages/Admin/assignment/AssignTable.js
+++ b/frontend/front/src/pages/Admin/assignment/AssignTable.js
@@ -29,10 +29,6 @@ const AssignTable = () => {
   const apiRef = useGridApiRef();
 
   const [selectedSurveyor, setSelectedSurveyor] = useState("");
-  const [paginationModel, setPaginationModel] = useState({
-    page: 0,
-    pageSize: 25,
-  });
 
   // Event handlers
   const handleChange = (event) => {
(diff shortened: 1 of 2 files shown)


lfilmeyer:
@Andrewy-gh Sorry if this issue has been confusing because I haven't been describing it properly - I don't think the issue was with the admin Assignments page (although maybe there was a different issue I didn't know about?)

The issue was with going to /surveyor/dashboard as a surveyor with lots of assignments. The more assignments the slower the page was when toggling open and shut the different assignments.

I tested viewing as a surveyor:
* 2 assignments - performance seems totally fine
* 29 assignments - (this is possibly how many assignments the surveyors will actually see, given we're expecting about 10 surveyors paired up across 145 assignments) - performance was noticeably laggy, but not show-stopping
* 145 assignments - performance was very very laggy

PR #700, opened 2025-06-24
Issue 656 frontend add newrelic

Andrewy-gh:
- Add logging of user language preference
- Add logging of visits to `/public/survey` and completion of public survey


mzagaja (approved):
Looks good, only question I have is do the env vars actually hide anything if we're exposing the values in index.js?

PR #707, opened 2025-08-19
Replace visit order with home id

Andrewy-gh:
- Replace visit order with home id for easier referencing
- Add a home id column for each home row in the assignment admin dashboard at `/admin/assignment/assignProfile/$id`


lfilmeyer:
Hey! Is this in response to this comment on slack: https://cfb-public.slack.com/archives/C03QW1GTJHW/p1755204901055929 ? Or was this something different? If yes in response, the tricky thing is that from the surveyor's perspective (not admin dashboard), the assignment IDs are listed 1 to N, which doesn't give them as helpful an identifier to pass on to one of the developers

PR #721, opened 2025-10-29
added option for Iphone users to select between apple maps and google maps

mcarre20:
-  created separate functions to generate link for apple maps and google maps
- created components for to open a modal for Iphone users the will allow them to select between Google Mpas or Apple Maps
- Android users route will always open on Google Maps.

### DIFF
--- a/frontend/front/src/pages/Surveyor/dashboard/AssignmentUnit.js
+++ b/frontend/front/src/pages/Surveyor/dashboard/AssignmentUnit.js
@@ -58,7 +58,9 @@ const AssignmentUnit = ({ data }) => {
   }, []);
 
   const handleGoogleMapBtn = () => {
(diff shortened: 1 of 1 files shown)


frontend/front/src/pages/Surveyor/dashboard/AssignmentUnit.js
```diff
@@ -57,20 +57,29 @@ const AssignmentUnit = ({ data }) => {
     setShowOnlyIncomplete((prev) => !prev);
   }, []);
 
+  const handleGoogleMapBtn = () => {
+    const sortedLocations = checked.map((id) => data.find((d) => d.id === id));
```
Andrewy-gh: Should `.sort` be chained as well? Since `sortedMaplocation` is doing that too, should it be changed to this?```const sortedLocations = checked
  .map((id) => data.find((d) => d.id === id))
  .sort(compareArrayByOrder);```

PR #728, opened 2026-01-06
Fix lang pref

Andrewy-gh:
- use getBrowserLanguageOrDefault() as fallback for LangPrefDropdown
- split up functions in lang urils


mzagaja (approved):
Let's do it!

PR #734, opened 2026-05-06
Update link in Partners.js to new URL

aTegart:
Update learn more link in footer to point to about us page instead of mass save


mzagaja (approved):
Solid sleuthing.

